PR #23, opened 2016-04-14
Android build on travis

samgiles:
Trying travis.

### DIFF
--- a/build-scripts/bootstrap.sh
+++ b/build-scripts/bootstrap.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -e
+
+echo "Installing dependencies from brew.."
+brew reinstall gradle flow xctool
+
+echo "Installing react-native.."
+npm install -g react-native-cli
+
+# Install project
+npm install
+
+if [[ "$BUILD_TYPE" == "android" ]];then
+    echo "bootstrapping android build.."
+    ./build-scripts/android/bootstrap.sh
+elif [[ "$BUILD_TYPE" == "ios" ]]; then
+    echo "TODO: bootstrap IOS build"
+else
+    echo "No BUILD_TYPE specified"
+    exit -1
+fi
--- a/build-scripts/build.sh
+++ b/build-scripts/build.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+if [[ "$BUILD_TYPE" == "android" ]];then
+    cd android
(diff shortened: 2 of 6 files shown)


wilsonpage:
Loookin' gouuuud! Any reason why we're not using `react-native` to build?

samgiles:
Because I haven't got that far yet, but I _think_ react native has put everything necessary in the gradle build config anyway.

wilsonpage:
Sick as fuck :sunglasses:

samgiles:
r? @wilsonpage 
Just builds an android release APK, doesn't do anything with it yet, just asserts that we can and nothing is broken.

We need an Apple developer account so we can code sign the iOS release, so I've disabled that for now.

build-scripts/bootstrap.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+set -e
+
+echo "Installing dependencies from brew.."
+# Install tools
+brew reinstall gradle flow watchman xctool
```
wilsonpage: I don't think `watchman` is required for production builds. Just monitoring JS files for changes and re-bundling when in development.

build-scripts/bootstrap.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "Installing react-native.."
+# Install react-native
+npm install -g react-native-cli
```
wilsonpage, line 8: Where is this used?

build-scripts/build.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+
+set -o pipefail
+
+if [[ "$BUILD_TYPE" == "android" ]];then
+    cd android
+    ./gradlew assembleRelease | tee
```
wilsonpage, line 7: Does this spit out the `.apk` somewhere?

build-scripts/bootstrap.sh
samgiles, line 8: The tools are used internally by the build for each platform. Running `./gradlew assembleRelease` or the equivalent `xcodebuild` command, fails without these tools available.

build-scripts/build.sh
samgiles, line 7: Yup, you end up with "magnet/android/app/build/outputs/apk/*.apk" after this PR I'll get them pushed to Github (or some other storage).

wilsonpage:
@samgiles let's ship it then. Feel free to land :)

PR #25, opened 2016-04-15
Added ability to fetch and parse locally (closes #22)

### LINKED ISSUE #22: Add ability to parse internal LAN URLs
Right now we depend on a cloud based metadata service. This service in unable to access services discovered on an internal LAN. Therefore we must have logic on the client-side to be able to fetch and parse in this scenario.

wilsonpage:


wilsonpage:
Apparently startups don't have time for reviews :wink: 

:ship: :ship: :ship: :dash:

PR #31, opened 2016-04-19
Ios support

arcturus:
@wilsonpage first r?

Big things here, and this is a discussion we already have :P

I created a new event to notify not just of btle beacon devices, but to notify about url.

The reason behind this, is that despite we dont' share that much code, later on we could potentially reuse this native files into other projects without any problem. Anyway, I agree that's arguable and I liked your point about embracing react.


index.ios.js
```diff
@@ -1,13 +1 @@
...
-
-import App from './lib/app';
-import { AppRegistry } from 'react-native';
-
-class Root extends App {
-
-}
-
-AppRegistry.registerComponent('Magnet', () => Root);
+require('./index.android.js');
```
wilsonpage, line 1: This seems a bit hacky. I'd rather abstract an `<App>` component and have that used by both `index.android.js` and `index.ios.js`.

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,204 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+
+/* ==!== Heavily edited for the URL discovery use case ==!== */
+
+import UIKit
```
wilsonpage: Do we need this?

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,204 @@
...
+  override init() {
+    super.init()
+    
+    self.centralManager = CBCentralManager(delegate: self, queue: self.beaconOperationsQueue)
+    self.centralManager.delegate = self
+    
+  }
+  
+  func startScanning() {
+    dispatch_async(self.beaconOperationsQueue) {
```
wilsonpage, line 68: What's `dispatch_async`? #lazyweb

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,204 @@
...
+  //TODO: add support for the URL eviction
+  private func removeURL(url: NSURL) {
+    self._urlFound.remove(url)
+  }
+}
+
+private func isAdvertisingURL(advertisementFrameList: [NSObject : AnyObject]) -> Bool {
+  let uuid = CBUUID(string: "FEAA")
+  guard let frameData = advertisementFrameList[uuid] as? NSData else { return false }
+  guard frameData.length > 1 else { return false }
```
wilsonpage: `guard` seems cool :)

index.ios.js
arcturus, line 1: I need to check if we can just include `index.js`.

lib/scanner/ble/index.js
```diff
@@ -32,11 +33,23 @@ class ScannerBle {
     this.callback(item.url);
   }
 
+  onURLFound(urls) {
```
wilsonpage: Nit: Camel-case

lib/scanner/ble/index.js
```diff
@@ -32,11 +33,23 @@ class ScannerBle {
...
+    debug('found urls', urls);
+    if (Array.isArray(urls)) {
+      urls.forEach((url) => {
+        this.callback(url);
+      });
+    }
+
```
wilsonpage: Nit: Remove line

lib/scanner/network/index.ios.js
```diff
@@ -0,0 +1,15 @@
...
+  start() {
+
+  }
+
+  stop() {
+
+  }
+}
+
+module.exports = NetworkScanner;
```
wilsonpage, line 15: Do we need separate files if both backends conform to the same API?

index.ios.js
wilsonpage, line 1: IIRC you must have a root `ios` and root `android` entry point.

ios/BeaconScanner.swift
arcturus, line 68: I think I commented, but didn't save it, dispatch_async will run the task off the main thread

lib/scanner/network/index.ios.js
arcturus, line 15: Yes, to avoid execution of unexisting native module

arcturus:
@wilsonpage 2nd r?

index.ios.js
wilsonpage, line 1: I don't mind refactoring this as a followup

ios/BeaconScanner.swift
wilsonpage, line 68: Where does that first arg (`self.beaconOperationsQueue`) come from? Just interested :)

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,225 @@
...
+  private func startScanningSynchronized() {
+    dispatch_async(self.beaconOperationsQueue) {
+      if self.centralManager.state != CBCentralManagerState.PoweredOn {
+        NSLog("CentralManager state is %d, cannot start scan", self.centralManager.state.rawValue)
+        self.shouldBeScanning = true
+      } else {
+        NSLog("Starting to scan for Eddystones")
+        let services = [CBUUID(string: "FEAA"), CBUUID(string: "FED8")]
+        let options = [CBCentralManagerScanOptionAllowDuplicatesKey : true]
+        self.centralManager.scanForPeripheralsWithServices(services, options: options)
```
wilsonpage, line 105: The function above is called `centralManager`. I'm confused where `.scanForPeripheralsWithServices` is defined.

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,225 @@
...
+      }
+    }
+  }
+  
+  private func addURL(url: NSURL) {
+    let count = self._urlFound.count
+    self._urlFound.insert(url)
+    if self._urlFound.count != count {
```
wilsonpage, line 113: I like this `if` block syntax :)

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,225 @@
...
+  } else {
+    return data?.length > 0
+  }
+  
+}
+
+private func URLForFrame(advertisementFrameList: [NSObject : AnyObject]) -> NSURL? {
+  let uribeaconUUID = CBUUID(string: "FED8")
+  var data = advertisementFrameList[uribeaconUUID] as? NSData
+  if (data == nil) {
```
wilsonpage, line 146: IMO it would be cleaner to `return parseBeacon(data)` here if we have the `data`.

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,225 @@
...
+  private func removeURL(url: NSURL) {
+    self._urlFound.remove(url)
+  }
+}
+
+private func isAdvertisingURL(advertisementFrameList: [NSObject : AnyObject]) -> Bool {
+  let eddystoneUUID = CBUUID(string: "FEAA")
+  let uribeaconUUID = CBUUID(string: "FED8")
+  let data = advertisementFrameList[uribeaconUUID]
+  if (data == nil) {
```
wilsonpage, line 128: IMO it would be cleaner to return here if we have the data :)

ios/BeaconScanner.swift
```diff
@@ -0,0 +1,225 @@
...
+    guard let frameData = advertisementFrameList[eddystoneUUID] as? NSData else { return false }
+    guard frameData.length > 1 else { return false }
+    
+    let count = frameData.length
+    var frameBytes = [UInt8](count: count, repeatedValue: 0)
+    frameData.getBytes(&frameBytes, length: count)
+    
+    return frameBytes[0] == 0x10
+  } else {
+    return data?.length > 0
```
wilsonpage, line 138: Remind me what the `?` means? :)

ios/magnet/ScannerBle.swift
```diff
@@ -0,0 +1,40 @@
...
+
+@objc(ScannerBle)
+class ScannerBle: NSObject, BeaconScannerDelegate {
+  
+  var bridge: RCTBridge!
+  var scanner: BeaconScanner!
+  
+  @objc func start() -> Void {
+    NSLog("Starting btle scanner");
+    self.scanner = BeaconScanner();
```
wilsonpage, line 19: So this is like `new BeaconScanner()`?

ios/magnet/ScannerBle.swift
```diff
@@ -0,0 +1,40 @@
...
+    self.scanner.delegate = self
+    self.scanner.startScanning();
```
wilsonpage: NIT: `self.scanner.start();` reads nicer IMO :)

ios/magnet/ScannerBle.swift
```diff
@@ -0,0 +1,40 @@
...
+  }
+  
+  func notify(urls: Array<String>) {
+    NSLog("Notifying JS context for the following urls");
+    for url in urls {
+      NSLog("%@", url);
+    }
+    self.bridge.eventDispatcher.sendDeviceEventWithName("magnet:urlfound", body: urls);
```
wilsonpage, line 29: Where is `self.bridge` defined? Line 14?
If so is `self.scanner` defined twice (ln 15 and ln 21)?

wilsonpage:
Some questions and NITs. Conditional r+ 😄

wilsonpage:
Also I'm unable to test without an iOS device. Can we order one today?

wilsonpage:
Also what's the situation on developer certificates. What's our team workflow? @samgiles do you know?

ios/magnet/ScannerBle.swift
arcturus, line 19: yes

ios/BeaconScanner.swift
arcturus, line 138: Means it's not unwrapped, can be the nsdata object or null, here we know it's not null.

ios/BeaconScanner.swift
arcturus, line 128: First we check for one kind of beacon and if not present we look for the other one

index.ios.js
arcturus, line 1: Then i prefer to leave it as it is, the meaning is clear, same entry point for ios and android :P

ios/BeaconScanner.swift
arcturus, line 105: `centralManager` is an object of type `CBCentralManager` https://developer.apple.com/library/ios/documentation/CoreBluetooth/Reference/CBCentralManager_Class

ios/BeaconScanner.swift
arcturus, line 146: Same than before we need to check with both uuid, for uribeacon and for eddystone

samgiles:
@wilsonpage It's in progress. Will keep the thread on magnet@moz up to date with progress.

PR #39, opened 2016-04-24
Fixing network error in ios

### LINKED ISSUE #34: Red screen of death: Network request failed
In iOS, and don't know if android as well, when we don't have connectivity or it drops, we show a red screen of death with a horrible react native message.

arcturus:
Fixes #34

### DIFF
--- a/lib/scanner/index.js
+++ b/lib/scanner/index.js
@@ -14,11 +14,11 @@ const SCAN_LENGTH = 10000; // 10secs
 var counter = 0;
 
 class Scanner {
-  constructor(callback) {
+  constructor(callback, netErrorListener = null) {
     this.items = {};
     this.callback = callback;
     this.onUrlFound = this.onUrlFound.bind(this);
-    this.netErrorListener = null;
+    this.netErrorListener = netErrorListener;
 
     this.scanners = [
       new NetworkScanner(this.onUrlFound),
@@ -159,16 +159,11 @@ class Scanner {
 
       .catch(err => {
         console.info(err);
-        if (this.netErrorListener) {
+        if (['NetworkError', 'TypeError'].indexOf(err.name) 
+          && this.netErrorListener) {
           this.netErrorListener(err);
         }
       });
-  }
-
-  set networkErrorListener(cb) {
-    if (typeof cb === 'function' || cb === null) {
-      this.netErrorListener = cb;
-    }
   }
 }
 
--- a/lib/scanner/index.js
+++ b/lib/scanner/index.js
@@ -18,6 +18,7 @@ class Scanner {
     this.items = {};
     this.callback = callback;
     this.onUrlFound = this.onUrlFound.bind(this);
+    this.netErrorListener = null;
 
     this.scanners = [
       new NetworkScanner(this.onUrlFound),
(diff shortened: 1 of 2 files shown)


arcturus:
I've discovered that `console.error` in iOS shows you the read screen of death, thing that doesn't happen in Android.

I've just replaced the console.error for console.info, and also setup a listener for displaying the error message on the visual component using the scanner.

@wilsonpage will you take a look? Do you like the prompt with the info or do you prefer a silent action?

lib/scanner/index.js
```diff
@@ -156,7 +157,18 @@ class Scanner {
         this.callback(data);
       })
 
-      .catch(err => console.error(err));
+      .catch(err => {
+        console.info(err);
+        if (this.netErrorListener) {
+          this.netErrorListener(err);
```
wilsonpage, line 167: Should we check if the error is network related? Currently this could be any error.

lib/scanner/index.js
```diff
@@ -156,7 +157,18 @@ class Scanner {
...
+        }
+      });
+  }
+
+  set networkErrorListener(cb) {
```
wilsonpage: Never seen this syntax before. Is there a reason for it?

lib/views/list.js
```diff
@@ -30,6 +31,9 @@ class List extends Component {
 
     this.onItemLongPress = this.onItemLongPress.bind(this);
     this.scanner = new Scanner(this.onItemFound.bind(this));
+    this.scanner.networkErrorListener = () => {
+      Alert.alert('Network error', 'Please check your internet connection');
+    };
```
wilsonpage: For consistency we should probably pass all callbacks into the constructor.

``` js
new Scanner({
  onItemFound: ...,
  onNetworkError: ...
});
```

lib/scanner/index.js
wilsonpage: Ah I get it now. See [comment](https://github.com/mozilla-magnet/magnet/pull/39/files#r60883335)

lib/views/list.js
arcturus: Makes sense

lib/scanner/index.js
arcturus, line 167: Yes, that's a good point, let's check for the network error

lib/scanner/index.js
```diff
@@ -156,7 +157,13 @@ class Scanner {
         this.callback(data);
       })
 
-      .catch(err => console.error(err));
+      .catch(err => {
+        console.info(err);
+        if (['NetworkError', 'TypeError'].indexOf(err.name)
```
arcturus: We are also receiving a TypeError: Network error message in some cases, so I added both to the list of error problems.

lib/views/list.js
```diff
@@ -29,7 +30,9 @@ class List extends Component {
     };
 
     this.onItemLongPress = this.onItemLongPress.bind(this);
-    this.scanner = new Scanner(this.onItemFound.bind(this));
+    this.scanner = new Scanner(this.onItemFound.bind(this), () => {
+      Alert.alert('Network error', 'Please check your internet connection');
+    });
```
wilsonpage, line 38: NIT: Would prefer this as a named options object. Otherwise I don't know what the second argument is for without reading the implementation of `Scanner`.

lib/scanner/index.js
wilsonpage: Are you missing `> -1` ?

PR #45, opened 2016-04-26
Deploy nightlies from travis when a nightly tag is created

samgiles:
r? @arcturus

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ before_install:
     - nvm use --delete-prefix 5.0
     - node --version && npm --version
 install:
-    - . build-scripts/bootstrap.sh
+    - source build-scripts/bootstrap.sh
 script:
     - ./build-scripts/build.sh
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,13 @@
 language: objective-c
+cache:
+    directories:
+#       - $HOME/.gradle/caches/
+#       - $HOME/.gradle/wrapper/
+        - node_modules/
 env:
-    - BUILD_TYPE=android ANDROID_HOME=/usr/local/opt/android-sdk
-    - BUILD_TYPE=ios
+    - BUILD_TYPE=android
+# Be a good travis citizen while we don't actually run the build on iOS yet.
+#   - BUILD_TYPE=ios
 before_install:
     - brew update
     - git clone https://github.com/creationix/nvm.git /tmp/.nvm
@@ -10,6 +16,8 @@ before_install:
     - nvm use --delete-prefix 5.0
     - node --version && npm --version
 install:
-    - ./build-scripts/bootstrap.sh
+    - . build-scripts/bootstrap.sh
 script:
     - ./build-scripts/build.sh
+after_success:
+    - ./build-scripts/deploy.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: objective-c
 cache:
+    skip_bundle_clean: true
     directories:
-#       - $HOME/.gradle/caches/
-#       - $HOME/.gradle/wrapper/
+        - $HOME/.gradle/caches/
+        - $HOME/.gradle/wrapper/
         - node_modules/
 env:
     - BUILD_TYPE=android
--- a/build-scripts/android/bootstrap.sh
+++ b/build-scripts/android/bootstrap.sh
@@ -3,7 +3,7 @@
 brew reinstall android-sdk
 
 export ANDROID_HOME=$(brew --prefix android-sdk)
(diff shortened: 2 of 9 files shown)


samgiles:
I've updated this so that the APK created doesn't need to be signed (it's a debug build)

.travis.yml
```diff
@@ -10,6 +16,8 @@ before_install:
     - nvm use --delete-prefix 5.0
     - node --version && npm --version
 install:
-    - ./build-scripts/bootstrap.sh
+    - . build-scripts/bootstrap.sh
```
arcturus: No need for `/`?

build-scripts/deploy-github.js
```diff
@@ -0,0 +1,176 @@
...
+              isPreRelease: true,
+        });
+    }
+
+    return release;
+  }).then((release) => {
+      return uploadFileToRelease(release, releaseInfo.binaryName, releaseInfo.path);
+  }).catch((e) => {
+      setTimeout(function() {
+          throw e;
```
arcturus, line 76: Why do we throw the error inside the setTimeout?

build-scripts/deploy-github.js
```diff
@@ -0,0 +1,176 @@
...
+            'Authorization': `Token ${githubAuthToken}`,
+            'User-Agent': userAgent,
+            'Content-Type': 'application/octet-stream'
+          },
+          formData: {
+            file: fs.createReadStream(artifactPath)
+          }
+        }, function(err, res, body) {
+            if (err) {
+                reject(err);
```
arcturus: nit:
`return reject(err);` or `reject(err); return;`

build-scripts/deploy-github.js
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            if (res.statusCode < 200 || res.statusCode > 299) {
+                reject(body);
```
arcturus: nit: same than above

build-scripts/deploy-github.js
```diff
@@ -0,0 +1,176 @@
...
+                'User-Agent': userAgent
+            },
+            json: {
+                tag_name: options.tag,
+                name: options.name,
+                prerelease: options.isPreRelease,
+            }
+        }, function(err, res, body) {
+            if (err) {
+                reject(err);
```
arcturus: and here

build-scripts/deploy-github.js
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            console.log(body);
+
+            if (res.statusCode < 200 || res.statusCode > 299) {
+                reject(body);
```
arcturus: Last one (sorry)

build-scripts/deploy-github.js
```diff
@@ -0,0 +1,176 @@
...
+    return new Promise(function(resolve, reject) {
+        console.log("Looking for release for tag: ", tag);
+        request.get({
+            url: getReleaseInfoUrl(tag),
+            headers: {
+                'User-Agent': userAgent,
+            }
+        }, function(err, res, body) {
+            if (err) {
+                reject(err);
```
arcturus: ;)

arcturus:
Looking good to me r+!

Left some minor comments.

build-scripts/deploy-github.js
samgiles, line 76: Because: `Promise.reject(0).catch((e) => { throw e; })` results in rejected `Promise` (because throw was caught in the catch block), by throwing in a `setTimeout` we throw on a task run on the main event loop and we break out of the Promise chain.

It's a common implementation of `done` or `finally` for a `Promise` to avoid errors that are otherwise lost on the promise chain.

Not sure if there is an alternative way of doing this since I last looked tbf.

arcturus, line 76: Pretty clear with the explanation :)

.travis.yml
samgiles: It's a `.` to source the file, I've updated it to `source` so it's clearer.

samgiles:
@arcturus r? again, I added release signing :)

build-scripts/android/bootstrap.sh
```diff
@@ -12,3 +12,13 @@ android=${ANDROID_HOME}/tools/android
 
 # Ensure android packages are installed
 ./build-scripts/android/accept-license.expect "./build-scripts/android/install-android-packages.sh"
+
+# Create APK signing variables
+
+# Get the release keystore from the keystore service
+wget https://${TOKEN_SERVICE_USER}:${TOKEN_SERVICE_AUTH_TOKEN}@ci.tengam.org/keystore/magnet-release.keystore
```
arcturus, line 19: Brilliant, simple and effective.

arcturus:
r+ shiiiiiip it!

Great work @samgiles

PR #61, opened 2016-05-02
No Bug - Added fastlane toolchain for iOS build automation

sleroux:
Hey!

I'm on the Firefox for iOS team and heard you were looking for some ways to package up the application for distribution for testers. For Firefox, we use Fastlane (https://github.com/fastlane) for our build automation. I believe it also supports Android if you're looking to automation that process down the road as well. I've setup a basic fastlane script that will build/package/upload the Magnet app to TestFlight so you can release the build to either internal/external users. 

If you're looking for more information about the build process for iOS feel free to reach out. For reference, here is the repo where we have our build automation tooling setup [1]

[1] https://github.com/mozilla/firefox-ios-build-tools

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ npm-debug.log
 
 # Never commit a keystore
 *.keystore
+
+# Fastlane
+ios/fastlane/report.xml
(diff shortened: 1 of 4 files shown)


samgiles:
Thanks, I'll give fastlane a go for Android soon.

PR #63, opened 2016-05-03
Use org.mozilla.magnet namespace

samgiles:
Chosen for consistency with org.mozilla.gecko ¯_(ツ)_/¯.  And we shouldn't use 'com.magnet', as someone else has almost certainly got that namespace and package names should be unique.

I think it's OK to not include 'android' in the namespace as we have done for iOS.

https://developer.android.com/guide/topics/manifest/manifest-element.html#package 

Note:

> Once you publish your application, **you cannot change the package name**. The package name defines your application's identity, so if you change it, then it is considered to be a different application and users of the previous version cannot update to the new version.

r? @wilsonpage

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -81,7 +81,7 @@ android {
     buildToolsVersion "23.0.1"
 
     defaultConfig {
-        applicationId "com.magnet"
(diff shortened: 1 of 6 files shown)


wilsonpage:
r+

PR #64, opened 2016-05-04
Don't remove all the tiles on refresh

samgiles:
Instead update the existing list and evict URLs we haven't seen for a while.  This means we don't need to re-render the whole list on refresh. :)

r? @wilsonpage

### DIFF
--- a/lib/scanner/index.js
+++ b/lib/scanner/index.js
@@ -135,28 +135,53 @@ class Scanner {
    * @public
    */
   clear() {
-    this.items = {};
+    const thirtySeconds = 1000 * 30;
+    const now = Date.now();
+
+    const keysToRemove = Object.keys(this.items).filter((key) => {
+      return (now - this.items[key].lastSeen) > thirtySeconds;
+    });
+
+    // Only update the view if there is something to change
+    let keysDeleted = false;
+    keysToRemove.forEach((key) => {
+      keysDeleted = true;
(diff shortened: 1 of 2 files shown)


lib/scanner/index.js
```diff
@@ -135,28 +135,53 @@ class Scanner {
...
   clear() {
-    this.items = {};
+    const thirtySeconds = 1000 * 30;
+    const now = Date.now();
+
+    const keysToRemove = Object.keys(this.items).filter((key) => {
+      return (now - this.items[key].lastSeen) > thirtySeconds;
+    });
+
+    // Only update the view if there is something to change
```
wilsonpage: NIT: Might be simpler to:

``` js
if (!keysToRemove.length) return;
```

lib/scanner/index.js
```diff
@@ -135,28 +135,53 @@ class Scanner {
    * @public
    */
   clear() {
-    this.items = {};
+    const thirtySeconds = 1000 * 30;
```
wilsonpage: Maybe have this as a `ITEM_EXPIRES` constant at the top of the file?

lib/scanner/index.js
wilsonpage: NIT: We could also to the work with one loop.

wilsonpage:
Nice work. R+ after a few mini NITs 👍

PR #66, opened 2016-05-07
Creating a padding for ios app

### LINKED ISSUE #52: Change ios status bar type
Status bar shouldn't overlay app content.

arcturus:
@wilsonpage decided to hide the status bar for both android and ios. Do you agree?

Fixes #52

### DIFF
--- a/index.android.js
+++ b/index.android.js
@@ -1,34 +1,18 @@
-/**
- * Sample React Native App
- * https://github.com/facebook/react-native
- */
-
 var ListView = require('./lib/views/list');
 var React = require('react-native');
 
 var {
-  AppRegistry,
   StyleSheet,
   Component,
   UIManager,
-  Platform,
-  View,
-  StatusBar
+  AppRegistry
 } = React;
 
 class App extends Component {
   render() {
-    return (<View>
-        <StatusBar
-          hidden={statusBar.hidden}/>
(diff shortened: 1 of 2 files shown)


wilsonpage:
I haven't flashed this patch yet, but I didn't think the intention was to hide the status bar. That kind of feature is usually reserved for immersive media experiences like video or games. I though we just want to make sure the ios status bar occupied space instead of floating on top of the app content.

If this isn't possible we could add some padding to the top of the app in IOS only.

arcturus:
Not hiding the status bar any more. Just added a padding.
Had to split the index into two as it's getting complicated reusing.

index.android.js
```diff
@@ -1,22 +1,16 @@
...
   Component,
   UIManager,
-  Platform
+  AppRegistry
 } = React;
 
 class App extends Component {
   render() {
-    return <ListView style={styles.list}/>;
+    return  <ListView style={styles.list}/>;
```
wilsonpage, line 13: Remove extra whitespace

wilsonpage:
@samgiles we've got a Travis failure:

`https://:@ci.tengam.org/keystore/magnet-release.keystore: Invalid user name.`

I don't think it's related to this patch, so going ahead and landing any way.

Issue #2, opened 2016-03-31
Add a way to dismiss tiles


wilsonpage:
When content is not relevant there should be a way for users to dismiss tiles, this decision should likely be persisted with the option to revoke if need be.

A swipe-away gesture might be nice. Although I'm concerned this might be tricky if swipe gesture conflicts with content inside a WebView tile. For example, if a tile has a horizontal range-slider component, how do we make sure the tile isn't mistakenly panned away during interaction?

This also raises the question of how touch events work inside a WebView inside a ScrollView. How do we prevent the ScrollView scrolling when the user is interacting with a touch based control?

Perhaps @etiennesegonzac knows more? :)

wilsonpage:
Let's wait for UX input before implementing anything here :)

samgiles:
We already have a context menu (which pops up when I'm refreshing sometimes) - maybe something like: https://github.com/dancormier/react-native-swipeout would be nice?

PR #76, opened 2016-05-11
Implement a way to dismiss tiles

wilsonpage:


wilsonpage:
@arcturus f?

This patch isn't a finished feature. I've done an initial pass at the UI interactions, but not yet sure if/how theses choices should be persisted somewhere.

With this patch you are able to swipe tiles left/right out of the list, the list with then compact and fill the space where the tile was.

Outstanding:
- Dismissed tiles are not persisted. If they were, we're need a way to remove them from the local blacklist. Perhaps as an initial pass we just make siping a 'fun' proof of concept?
- Scroll position of list needs to correct itself in some situations once content height has changed
- The swipe touch handlers conflict with the context menu feature we have. I'd suggest we remove the context menu feature in anticipation of the next expanded/detail/interactive tile view we're planning. We cannot have both as the context menu could pop up mid swipe, tricky heuristics etc.

lib/views/app.js
```diff
@@ -97,10 +98,33 @@ class AppView extends React.Component {
...
+  setItems(items) {
+    LayoutAnimation.configureNext(LayoutAnimation.Presets.easeInEaseOut);
+    this.setState({ items });
+  }
+
+  removeItem(id) {
+    debug('remove item', id);
+    var items = this.state.items;
+
+    for (let i = 0; i < items.length; i++) {
```
arcturus, line 123: Can we do `items.find((item) => {item.id === id})`

lib/views/list.js
```diff
@@ -65,9 +68,14 @@ class ListView extends React.Component {
...
     debug('render list items');
     return this.props.items.map(item => {
+      debug('render item', item.id);
       return <ListItem
         {...item}
         onLongPress={this.onItemLongPress}
+        onGestureStart={this.onItemGestureStart.bind(this)}
+        onGestureEnd={this.onItemGestureEnd.bind(this)}
+        onSwiped={this.onItemSwiped.bind(this)}
+        swipable={true}
```
arcturus: Why do we need the attribute `swipable` if it's contastant? Preparing for a follow up?

arcturus:
Can I be picky and ask for unit tests? ;)

wilsonpage:
> Can I be picky and ask for unit tests? ;)

Sure! This patch was just me working out how this could work. No excuses now we have the plumbing in place.

lib/views/list.js
wilsonpage: I added it in case we wanted to land this feature behind a flag. If we're happy to go ahead I'll remove the flag.

lib/views/app.js
wilsonpage, line 123: Sure. Although I was wary of support, if this is being run inside an old Android JS env. Does babel cover these kind of APIs?

ios/magnet/AppDelegate.m
```diff
@@ -31,7 +31,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
    * on the same Wi-Fi network.
    */
 
-//   jsCodeLocation = [NSURL URLWithString:@"http://localhost:8081/index.ios.bundle?platform=ios&dev=true"];
+   jsCodeLocation = [NSURL URLWithString:@"http://192.168.0.7:8081/index.ios.bundle?platform=ios&dev=true"];
```
arcturus: dont forget to revert this change

lib/views/list.js
arcturus: I like the idea of making this configurable, can we get the value from `config.js` ?

ios/magnet/AppDelegate.m
samgiles: I use `git add -p` to ignore stuff like this when adding and reviewing my changes before pushing - may or may not be useful.

wilsonpage: You're right. I'm la💤y :)

arcturus:
Fixes #2

PR #77, opened 2016-05-12
Rename stuff to Project Magnet

samgiles:
I think this is all the places.

I couldn't see where iOS picks up the name for the home screen, is it from the App ID I wonder? @arcturus ?

https://github.com/mozilla-magnet/magnet/issues/2


samgiles:
Ok - failing builds are useless, let's only do a full build on nightly and master merges.

android/Magnet.iml
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<module external.linked.project.id="Magnet" external.linked.project.path="$MODULE_DIR$" external.root.project.path="$MODULE_DIR$" external.system.id="GRADLE" external.system.module.group="" external.system.module.version="unspecified" type="JAVA_MODULE" version="4">
+<module external.linked.project.id="Project Magnet" external.linked.project.path="$MODULE_DIR$" external.root.project.path="$MODULE_DIR$" external.system.id="GRADLE" external.system.module.group="" external.system.module.version="unspecified" type="JAVA_MODULE" version="4">
```
wilsonpage: Are spaces ok?

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "Magnet",
+  "name": "project-magnet",
```
wilsonpage: `react-native` uses this string to configure the some of the Android and iOS applications. So we should make sure that it's valid in the contexts it may be used during a `react-native upgrade`.

wilsonpage:
@samgiles any progress on this?

package.json
samgiles: Yeh, I think we've already got a mismatch because we're using org.mozilla namespace.  But I'm not going to touch this as there is all sorts of weird stuff overwritten and changed when running `react-native upgrade`

samgiles:
Turns out 'iml' files should not be committed  - gradle should remain the source of truth.

Issue #81, opened 2016-05-13
Travis nightly builds failing


wilsonpage:
https://travis-ci.org/mozilla-magnet/magnet-client/builds/129906682

samgiles:
Hmm, this is why we need builds on PR.  Seems like a regression. But, they're not reliable enough :/

wilsonpage:
@samgiles can this be closed now?

samgiles:
No - they're no longer failing but I noticed they're not being uploaded anymore.

samgiles:
Ok, fixed now :)

PR #84, opened 2016-05-13
Remove bad Activity

samgiles:
@wilsonpage This was causing #81, not sure why it was needed if at all.


wilsonpage:
@samgiles ah, I think this was introduced by the recent `react-native upgrade`. I've removed this file once before, keeps finding its way back!

wilsonpage:
Let's wait till green and land?

wilsonpage:
@samgiles now we're getting that timeout:

```
No output has been received in the last 10m0s, this potentially
indicates a stalled build or something wrong with the build itself.
```

samgiles:
Let's merge this, the push check passed with Travis and no-one can really build off of a broken master.

Issue #83, opened 2016-05-13
Make Travis builds more reliable on pull request.


samgiles:
We get a lot of failures and timeouts, which means bad code might get merged - we need a way of improving this.

PR #87, opened 2016-05-13
[WIP] Try android specific builds

samgiles:
Working on #83


samgiles:
Builds take ~10 minutes now instead of 15.  This should make the builds a lot stabler and quicker to run.  It now uses the containerised build infrastructure too, so the builds should be quicker to start.

wilsonpage:
@samgiles I'll just have to trust you on this one. No idea what this code does :p

PR #107, opened 2016-05-23
Change webview link clicks open in device browser (android) (closes #105)

### LINKED ISSUE #105: Links inside webview should open in browser

wilsonpage:


wilsonpage:
@samgiles r?

android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebViewClient.java
```diff
@@ -22,24 +22,38 @@ public void onPageFinished(WebView webView, String url) {
         Handler handler = new Handler();
         handler.postDelayed(new Runnable() {
             public void run() {
+                magnetWebView.loaded = true;
```
wilsonpage: Is this a hacky pattern? #androidnoob

samgiles: Yeh it is a bit, it's kind of weird that the client sets the loaded property.

wilsonpage: What works better?

samgiles: Seems like just storing the state in the client makes sense.

android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebView.java
```diff
@@ -15,6 +15,7 @@
 public class MagnetWebView extends WebView implements LifecycleEventListener {
     String TAG = "MagnetWebView";
     private boolean layoutSet;
+    public boolean loaded = false;
```
samgiles: I think adding a public property, that's not used by the owner is a hint that something's up.  See comment below.

android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebView.java
```diff
@@ -15,6 +15,7 @@
 public class MagnetWebView extends WebView implements LifecycleEventListener {
     String TAG = "MagnetWebView";
```
samgiles: Let's take the opportunity to make this `private static final String TAG = "MagnetWebView"`.

samgiles:
With comments addressed LGTM!

wilsonpage:
@samgiles comments addressed. Landing on green.

Issue #71, opened 2016-05-10
Investigating Monkeyrunner for functional tests


isabelrios:
Started to investigate how to use monkeyrunner for android.
I got a simple test to install the app both in a real device and in the emulator.
But if the idea is to have the tests running on the emulator there is the problem that it does not support bluetooth, so we could only simulate a limited tests like to install the app, see the message that bluetooth is needed and unistall it.
The only option I see is to run the tests in a real device...but I would probably be missing more options :)
cc @samgiles

samgiles:
Hmm, that's an issue, we would miss out on testing the bluetooth discovery automatically, but, we could test the UI using mDNS though.

isabelrios:
Sorry, my first comment was not totally correct, I got to install the app in the emulator but it was closed a few seconds later with the message: Unfortuntely, magnet has stopped. Thought was something related to my emulator, but it is not, it is related to the blueetooth not being supported by the emulator.
I tried modifying a little bit the app, removing the blutooth permission and so and now I get the app open, it is not closed, but with errors of course (I already run adb reverse tcp:8081 tcp:8081).
![screenshot_20160513-151018](https://cloud.githubusercontent.com/assets/1897507/15250253/2b4bf3ca-1924-11e6-901a-e88672b0b86a.png)

So in case we want to have some tests on the emulator, all the bluetooth part has to be modified :S Not sure if we can do all using mDNS (sorry not very familiar with that) @samgiles

arcturus:
:( yeah, just read that the emulator doesnt support BT (among other things, like wifi)

http://developer.android.com/tools/devices/emulator.html#starting

samgiles:
@isabelrios Incidentally, there should  be a fix for that particular error you're seeing in #80 

We can add probably add to the config file on a build to have tiles show up by default. https://github.com/mozilla-magnet/magnet-client/blob/master/config.js#L23

isabelrios:
@samgiles Thanks! I will try again and see if with that at least we can open the app via emulator without crashes, then adding some tiles by default would definitely help to start testing the builds on the emulator.

isabelrios:
Hi @samgiles, some progress here. 
As you said, that error is not shown anymore. And, changing the conf file, I can see the defaults tiles on the emulator.
But, since the app keeps scanning, searching for beacons, after some seconds in which I see the default tiles
[...]

PR #113, opened 2016-05-25
Check for null reference on bluetoothAdapter.

samgiles:
Addresses the issue in #71.

r? @wilsonpage

### DIFF
--- a/android/app/src/main/java/org/mozilla/magnet/ScannerBle.java
+++ b/android/app/src/main/java/org/mozilla/magnet/ScannerBle.java
@@ -53,7 +53,8 @@ public class ScannerBle extends ReactContextBaseJavaModule implements ActivityEv
             return;
         }
 
-        if (!bluetoothAdapter.isEnabled()) {
+        // Try and start bluetooth
+        if (!isBluetoothEnabled()) {
             Intent enableBluetoothIntent = new Intent(BluetoothAdapter.ACTION_REQUEST_ENABLE);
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/org/mozilla/magnet/ScannerBle.java
```diff
@@ -66,6 +66,11 @@ public void start() {
     @ReactMethod
     public void stop() {
         Log.d(TAG, "stop");
+        if (bluetoothAdapter == null || !bluetoothAdapter.isEnabled()) {
+            // Device does not support bluetooth, or is not enabled
+            return;
+        }
+
```
wilsonpage, line 74: It might be cleaner to rely on the single check in `.start()` and set a `started` flag that we check for inside `.stop()`.

samgiles, line 74: Hmm, we'd still need to check for `isEnabled` in both places regardless.  Not certain it's entirely worth duplicating the initialisation state of `bluetoothAdapter`.

wilsonpage, line 74: k, how about we abstract the logic behind a `bluetoothEnabled()` method then?

``` java
if (!bluetoothEnabled()) {
  return;
}
```

wilsonpage:
Let's get this closed before the end of this sprint Sammy Boy 😄

PR #115, opened 2016-05-26
Implemented expanded view (closes #85)

### LINKED ISSUE #85: Implement 'expanded' view
We need a visual for what the expanded/detail view of the tile looks like.

wilsonpage:


wilsonpage:
@samgiles some more tweaks needed, but ready for an initial f?

Still to do: 
- [x] Check on iOS
- [x] Fix touches not working properly on webview content
- [x] Unit-tests

samgiles:
lgtm so far.  Do we need `var`s everywhere though?  Pretty sure a `const` will do in most cases here.

wilsonpage:
Just waiting on green. If we want to swap all `var` for `const` then we should do that in a separate issue as it's unrelated to this patch. Although I'm still not convinced it brings much value. Just seems like more chars to write IMO.

arcturus:
+1 to use const when makes sense ;) (sorry @wilsonpage )

wilsonpage:
@arcturus followup #120

PR #133, opened 2016-06-08
Remove the hamburger menu for now

samgiles:
@wilsonpage r?

### DIFF
--- a/lib/views/header-bar.js
+++ b/lib/views/header-bar.js
@@ -64,7 +64,6 @@ class HeaderBarView extends React.Component {
   renderLeft() {
     return <View style={styles.left}>
       {this.renderCloseButton()}
-      {this.renderHamburgerButton()}
     </View>
   }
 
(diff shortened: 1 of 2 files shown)


lib/views/header-bar.js
```diff
@@ -84,22 +83,6 @@ class HeaderBarView extends React.Component {
...
-    return <TouchableOpacity
-      testId="hamburger-button"
-      style={styles.hamburger}
-      onPress={this.props.onHamburgerPressed.bind(this)}>
-      <Image
-        style={styles.hamburgerImage}
-        source={require('../images/header-bar-hamburger.png')}/>
-    </TouchableOpacity>
-  }
-
```
wilsonpage, line 102: Can you keep this function so that we can easily add it back later.

wilsonpage:
A unit-test is failing. Can you `it.skip(` it?

wilsonpage:
r+ after comments addressed

lib/views/header-bar.js
samgiles, line 102: But it's in version control history! Not a fan of keeping dead code around fwiw.

PR #139, opened 2016-06-08
Grow overflowing webviews when tile expanded (closes #125)

### LINKED ISSUE #125: Grow overflowing webviews when 'expanded'
When an embed tile that is clamped by `MAX_HEIGHT` is expanded we should also grow the tile to fill the available space.

If the content is still to tall, the webview can scroll.

wilsonpage:


wilsonpage:
@samgiles r?

lib/views/list-item/content/embed.js
```diff
@@ -25,10 +25,10 @@ var {
  *
  * @type {Number}
  */
-const MAX_HEIGHT = 450;
+const MAX_HEIGHT = 460;
 
 /**
- * Default initial height used if we're
+ * The minimum height used if we're
```
samgiles: We should rename from `DEFAULT_HEIGHT` to `MIN_HEIGHT`.

lib/views/list-item/index.js
```diff
@@ -369,17 +376,13 @@ const styles = StyleSheet.create({
...
+    paddingBottom: MARGIN_HORIZONTAL,
 
-    // ios
-    shadowOffset: { width: 0, height: 0.7 },
-    shadowColor: '#000',
-    shadowOpacity: 0.25,
-    shadowRadius: 0,
+    // This seems to be required to make
+    // sure taps register on embed tiles,
+    // not sure why?! Needs deeper look.
```
samgiles, line 385: File a follow up issue, and reference here.

samgiles:
Nice! Just a couple of issues, r+ once they're addressed.  

We also need to follow up on the weird scroll position changing of the scroll view in the expanded view when the on screen keyboard is shown.

PR #148, opened 2016-06-09
We need to force a size changed event _after_ layout so that the chrome web view has the correct size.

samgiles:
@wilsonpage r? (Also check it works for you now :p)

Might also be worth checking this doesn't mess up the webview you were previously using.

### DIFF
--- a/android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebView.java
+++ b/android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebView.java
@@ -4,6 +4,8 @@ import android.os.Build;
 import android.util.Log;
 import android.webkit.WebView;
 
+import android.view.View;
+
 import com.facebook.react.bridge.LifecycleEventListener;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.WritableMap;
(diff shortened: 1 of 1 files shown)


android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebView.java
```diff
@@ -61,18 +68,39 @@ public void cleanupCallbacksAndDestroy() {
 
     @Override
     protected void onLayout(boolean changed, int left, int top, int right, int bottom) {
-        super.onLayout(changed, left, top, right, bottom);
         Log.d(TAG, "on layout");
         setLayout();
+
+        super.onLayout(changed, left, top, right, bottom);
```
wilsonpage, line 70: Any reason this isn't called before setLayout?

samgiles, line 70: No reason - I changed.

samgiles:
Sorry, I noticed loads of bad stuff after I r?'d it. 😬

wilsonpage:
Working on my device!!! \o/ r+

samgiles:
The travis job's output says they finished successfully... So I'm gonna merge.

PR #153, opened 2016-06-10
Fix tiles cannot be swiped (fixes #150)

### LINKED ISSUE #150: Tiles cannot be swiped off
It is possible to swipe a tile and it moves, but swipping it off does not work.

wilsonpage:


wilsonpage:
@arcturus r?

PR #157, opened 2016-06-11
Fixed open url in browser causing crash (fixes #156)

### LINKED ISSUE #156: Crash: Unfortunately, project magente has stopped
This has happened with today nightly build  eec3399:
1. Tapping on an event in a sched tile in the expanded view: https://mzl.la/schgw2 (Ideation Skill Build Intro)
2. Trying to do something like retweet from the twitter tile in the expanded view: https://mzl.la/twtldn

Only android build is affected.

wilsonpage:


wilsonpage:
@arcturus r?

I'm not really sure why this started throwing. It seems we needed to pass a special flag to the activity to make it work. Not sure if you have any Android knowledge in this area tat can shed any light.

arcturus:
Most likely a change in react native.

Anyway, launching a browsing activity when we navigate want to open the tile sounds good to me.

R+

PR #173, opened 2016-07-04
Make the app_name parameter in metrics fixed, so it's the same for io…

### LINKED ISSUE #172: [Metrics] Set appname to a constant value
Right now we get the appname paremeter as well from the operating system, having different values.

We will need to setup an unique and constant value for this in order to extract the data from the big table correctly.

arcturus:
…s and android.

Closes #172 

@wilsonpage r?


arcturus:
Merging after talking with @wilsonpage

PR #176, opened 2016-07-06
Implemented Android background notifications using magnet-scanner-android

wilsonpage:


wilsonpage:
@arcturus you hate me I know (r?)

ios/MagnetScanner.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  MagnetScanner.swift
+//  magnet
+//
+//  Created by Wilson Page on 05/07/2016.
+//  Copyright © 2016 Facebook. All rights reserved.
```
arcturus: This is not copyrighted by FB ;)

ios/MagnetScannerReact.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  ReactScanner.swift
+//  magnet
+//
+//  Created by Wilson Page on 05/07/2016.
+//  Copyright © 2016 Facebook. All rights reserved.
```
arcturus: same here

ios/ScannerBle.swift
```diff
@@ -89,7 +61,10 @@ class BeaconScanner: NSObject, CBCentralManagerDelegate {
     guard isAdvertisingURL(serviceData) else { return }
     
     if let url = URLForFrame(serviceData) {
-      self.addURL(url)
+      callback([
+        "url": url.absoluteString,
+        "distance": -1
```
arcturus, line 66: We could potentially calculate the distance right? Are we planning to do this in a follow up?

ios/ScannerBle.swift
```diff
@@ -18,50 +18,22 @@
...
-      for url in _urlFound {
-        urls.insert("\(url)");
-      }
-      return Array(urls);
-    }
-  }
+  private var callback: ((Dictionary<String, AnyObject>) -> Void)!
   
-  override init() {
+  init(callback: (Dictionary<String, AnyObject>) -> Void) {
```
arcturus, line 32: I liked the way that we were using before, defining a delegate so the core class needs to implements the delegate. Seems to me more ios than passing the callback function.

arcturus:
iOS changes looking good to me so far, the only comment is about implementing the delegate pattern.

lib/scanner/index.js
```diff
@@ -6,9 +6,15 @@
...
+const ReactNative = require('react-native');
 const metadata = require('./metadata');
-const BleScanner = require('./ble');
+
+const {
+  NativeModules,
+  DeviceEventEmitter
+} = ReactNative;
+
+const MagnetScanner = NativeModules.MagnetScannerReact;
```
arcturus, line 17: Love the simplicity of the code now 👍 you rock.

lib/views/app.js
```diff
@@ -112,7 +112,7 @@ class AppView extends React.Component {
           onItemSwiped={this.onItemSwiped.bind(this)}/>
         <HeaderBarView
           style={styles.headerBar}
-          scanning={this.state.scanning}
+          scanning={false}
```
arcturus: Do we need this property?

ios/ScannerBle.swift
wilsonpage, line 32: I did some research into this and came to the conclusion that callbacks are cleaner and more modern and delegates were an older pattern. Here's a [blog post](https://medium.com/cobe-mobile/why-you-shouldn-t-use-delegates-in-swift-7ef808a7f16b#.o5mhn7l5w) for more info.

Also [here's an example](https://developer.apple.com/library/prerelease/iOS/documentation/UIKit/Reference/UIAlertAction_Class/index.html#//apple_ref/occ/clm/UIAlertAction/actionWithTitle:style:handler:) of Apple using callbacks in a newer core library.

ios/ScannerBle.swift
wilsonpage, line 66: Yeah, that is a missing feature of the ios library and should be introduced as a followup

lib/views/app.js
wilsonpage: Ah, don't think we do now :)

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
```diff
@@ -0,0 +1,90 @@
...
+import com.facebook.react.bridge.ReactApplicationContext;
+import com.facebook.react.bridge.ReactContextBaseJavaModule;
+import com.facebook.react.bridge.ReactMethod;
+import com.facebook.react.bridge.WritableMap;
+import com.facebook.react.modules.core.DeviceEventManagerModule;
+
+import org.mozilla.magnet.net.scanner.MagnetScannerItem;
+
+public class MagnetScannerReact extends ReactContextBaseJavaModule implements ScannerService.ScannerServiceCallback {
+    String TAG = "MagnetScannerReact";
```
arcturus: this can be `private`

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
```diff
@@ -0,0 +1,90 @@
...
+    public void stop() {
+        if (!mBound) {
+            return;
+        }
+
+        mService.stop();
+    }
+
+    @Override
+    public void onItemFound(MagnetScannerItem item) {
```
arcturus, line 77: Should we add telemetry to this method?

Or do you prefer to do it in a follow up?

android/app/src/main/java/org/mozilla/magnet/NotificationService.java
```diff
@@ -0,0 +1,174 @@
...
+     * for a few seconds (SCAN_DURATION_MS). Once complete
+     * the service terminates itself, which in-turn triggers
+     * the cleanup in `onDestroy()`.
+     */
+    private void onConnected() {
+        mScannerService.addListener(NotificationService.this);
+        mItems = mScannerService.getItems();
+        updateNotification();
+
+        mScannerService.start();
```
arcturus, line 105: why do we call `start` after updating the notifications?

android/app/src/main/java/org/mozilla/magnet/ScannerService.java
```diff
@@ -0,0 +1,186 @@
...
+import org.mozilla.magnet.net.scanner.MagnetScannerCallback;
+import org.mozilla.magnet.net.scanner.MagnetScannerItem;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class ScannerService extends IntentService implements MagnetScannerCallback {
+    String TAG = "ScannerService";
```
arcturus: `private final static` ..

android/app/src/main/java/org/mozilla/magnet/ScannerService.java
```diff
@@ -0,0 +1,186 @@
...
+
+    /**
+     * A hash to store found mItems.
+     *
+     * Items expire after a short period of time.
+     */
+    private ExpiringMap<String, MagnetScannerItem> mItems = ExpiringMap.builder()
+            .expiration(30, TimeUnit.SECONDS)
+            .build();
+//    private HashMap<String, MagnetScannerItem> mItems = new HashMap();
```
arcturus: nit: dead code

android/app/src/main/java/org/mozilla/magnet/ScannerService.java
```diff
@@ -0,0 +1,186 @@
...
+     *
+     * @param item
+     */
+    @Override
+    public void onItemFound(MagnetScannerItem item) {
+        Log.d(TAG, "on item found: " + item.getUrl());
+        mItems.put(item.getUrl(), item);
+
+        for (ScannerServiceCallback callback : mCallbacks) {
+            callback.onItemFound(item);
```
arcturus, line 94: Should we notify always, or leave that to someone else?

android/app/src/main/java/org/mozilla/magnet/ScannerService.java
```diff
@@ -0,0 +1,186 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class ScannerService extends IntentService implements MagnetScannerCallback {
+    String TAG = "ScannerService";
+    private List<ScannerServiceCallback> mCallbacks = new ArrayList<>();
+    private LocalBinder mBinder = new LocalBinder();
+    private MagnetScanner mMagnetScanner;
+    private int mClients = 0;
```
arcturus, line 25: nit: mNumClients

ios/ScannerBle.swift
arcturus, line 32: Oh, great article, good to know.

arcturus:
Testing it out in ios, I realised that several times, when forcing the refresh, it gets stuck in the 'Nothing found' state. Not sure yet why.

arcturus:
Also we don't have a way of not notifying beacons that you find daily right?

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
wilsonpage, line 77: Should we do the telemetry in JS land where we only have to write once and already have the metrics libraries in place?

wilsonpage, line 77: But yeah follow up is possible ;)

android/app/src/main/java/org/mozilla/magnet/NotificationService.java
wilsonpage, line 105: We call `updateNotifications()` here to update the state of current notification. This may result in a stale notification being removed or if we got some items back from the service the notification will be updated.

Then if/when we find items, we will update again.

android/app/src/main/java/org/mozilla/magnet/ScannerService.java
wilsonpage, line 94: Not sure I follow

wilsonpage:
> Also we don't have a way of not notifying beacons that you find daily right?

We are not collecting a history yet. I'd suggest that that feature be part of this.

I think we should write an android and ios module for get/set history and blocked/dismissed tiles; we can expose this module to JS land so that we can respond to swiped tiles, but also make use of it natively in `NotificationService` to decide whether we should show notifications or not.

We could even cache some metadata so that we can show richer notifications without hitting the network.

wilsonpage:
Action items:
- [x] Check refresh is working correctly on iOS
- [x] Add a flag to disable notifications

wilsonpage:
@arcturus addressed above points. iOS seems to be working as expected on my device.

PR #181, opened 2016-07-12
Restored bluetooth prompt (fixes #174)

### LINKED ISSUE #174: [Notifications] Android support
Following discussions on #14 we will need to implement android notifications.

wilsonpage:


wilsonpage:
@arcturus r?

wilsonpage:
I've implemented as much of this on the JS side as possible. I found displaying alerts/prompts outside of ReactNative framework was messing with timings. I think we should have the rule that all UI related work should be controlled by the JS layer.

android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebViewManager.java
```diff
@@ -48,8 +48,6 @@ protected WebView createViewInstance(ThemedReactContext reactContext) {
         MagnetWebView magnetWebView = new MagnetWebView(mContext);
         reactContext.addLifecycleEventListener(magnetWebView);
 
-        Context otherContext = magnetWebView.getContext();
```
arcturus, line 51: What is this change? Is it related to this PR?

wilsonpage, line 51: Dead/unused variable I found.

arcturus:
Works perfect in android n too, shiiiip it!

r+!

android/app/src/main/java/org/mozilla/magnet/webview/MagnetWebViewManager.java
arcturus, line 51: Fair enough!

Issue #175, opened 2016-07-04
[Notifications] iOS support


arcturus:
Following discussions on #14 we will need to implement android notifications.

PR #186, opened 2016-07-13
Allowing background scanning and badge notifications when app is open

arcturus:
@wilsonpage r?

### DIFF
--- a/ios/MagnetScannerReact.swift
+++ b/ios/MagnetScannerReact.swift
@@ -3,9 +3,12 @@ import Foundation
 @objc(MagnetScannerReact) class MagnetScannerReact: NSObject {
   var scanner: MagnetScanner!;
   var bridge: RCTBridge!
+  var notifyTimer: NSTimer!
+  var toNotify: [String]!
   
   override init() {
     super.init()
+    toNotify = []
     scanner = MagnetScanner(callback: onItemFound)
   }
   
@@ -14,12 +17,49 @@ import Foundation
   }
   
   @objc func stop() {
-    scanner.stop()
+    // We don't stop scanning when the app goes into the background as we want bluetooth scanning to continue.
+    // If we stop scanning we have no way of starting again until the app comes to the foreground.
+    // When the app is in the background and the scanner is still running,
+    // the OS will periodically call the scan-callback with results.
   }
   
   func onItemFound(item: Dictionary<String, AnyObject>) {
-    print("item found", item["url"]);
+    print("item found", item["url"])
     bridge.eventDispatcher()
-      .sendDeviceEventWithName("magnetscanner:itemfound", body: item);
+      .sendDeviceEventWithName("magnetscanner:itemfound", body: item)
+    
+    notifyUser(item["url"] as! String)
+  }
+  
+  // Throttles the notification process, waiting for 1 minute until
+  // setting up the badge with the number of elements nearby.
+  func notifyUser(url: String) {
+    if (toNotify.contains(url)) {
+      return;
+    }
+    toNotify.append(url)
+    if (notifyTimer != nil) {
+      notifyTimer.invalidate();
+    }
+    
+    // We need to throttle the notifications, waiting for receive more nearby web pages,
+    // so we use NSTimer to wait for 1 minute and after that we trigger the action, in
+    // this case doNotifyUser, that notifies depending on the number of web pages found
+    // during the waiting period.
+    notifyTimer = NSTimer.init(
+      timeInterval: 60 ,
+      target: self,
+      selector: #selector(MagnetScannerReact.doNotifyUser),
+      userInfo: nil,
+      repeats: false)
+    // It's important to note that this timer is being execute in the main loop
+    // (look at NSRunLoop.mainRunLoop) to be sure that the notifications will reflect
+    // changes in the UI despite of this being trigget in the background.
+    NSRunLoop.mainRunLoop().addTimer(notifyTimer, forMode: NSDefaultRunLoopMode)
+  }
+  
+  func doNotifyUser() {
+    NotificationsHelper.updateBadge(toNotify.count);
+    toNotify = [];
   }
 }
--- a/ios/ScannerBle.swift
+++ b/ios/ScannerBle.swift
@@ -25,7 +25,9 @@ import CoreBluetooth
 ///
 class BeaconScanner: NSObject, Scanner, CBCentralManagerDelegate {
   private var centralManager: CBCentralManager!
-  private let beaconOperationsQueue: dispatch_queue_t = dispatch_queue_create("beacon_operations_queue", nil)
+  // As we are going to background we will need to dispatch our queue for attending BTLE operaions in a serial queue.
+  // more information: https://developer.apple.com/library/ios/documentation/General/Conceptual/ConcurrencyProgrammingGuide/OperationQueues/OperationQueues.html
+  private let beaconOperationsQueue: dispatch_queue_t = dispatch_queue_create("beacon_operations_queue", DISPATCH_QUEUE_SERIAL)
   private var shouldBeScanning: Bool = false
   private var callback: ((Dictionary<String, AnyObject>) -> Void)!
   
--- a/ios/magnet/AppDelegate.m
+++ b/ios/magnet/AppDelegate.m
@@ -11,6 +11,9 @@
 
 #import "RCTRootView.h"
 
+// Include the project headers for swift code. (<project-name>-Swift.h)
+#import "magnet-Swift.h"
+
 @implementation AppDelegate
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions
@@ -52,8 +55,19 @@
   UIViewController *rootViewController = [UIViewController new];
   rootViewController.view = rootView;
   self.window.rootViewController = rootViewController;
+  [NotificationsHelper register];
   [self.window makeKeyAndVisible];
   return YES;
+}
+
+- (void)applicationDidBecomeActive:(UIApplication *)application {
+  [NotificationsHelper clearNotifications];
+  [NotificationsHelper disable];
+}
+
+- (void)applicationDidEnterBackground:(UIApplication *)application {
+  [NotificationsHelper clearNotifications];
+  [NotificationsHelper enable];
 }
 
 @end
--- a/ios/NotificationsHelper.swift
+++ b/ios/NotificationsHelper.swift
@@ -1,0 +1,36 @@
+//
+//  NotificationsHelper.swift
+//  magnet
+//
+//  Created by Francisco Jordano on 12/07/2016.
+//
+
+import Foundation
+import UIKit
+
+@objc(NotificationsHelper) class NotificationsHelper: NSObject {
+  static var enabled: Bool = true
+  
+  @objc class func register() {
+    let notificationsSettings = UIUserNotificationSettings(forTypes: [.Badge], categories: nil)
+    UIApplication.sharedApplication().registerUserNotificationSettings(notificationsSettings)
+  }
+  
+  class func updateBadge(num: Int) {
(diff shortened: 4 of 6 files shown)


arcturus:
This pr closes #175

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,41 @@ import Foundation
   }
   
   @objc func stop() {
-    scanner.stop()
+    // We don't stop scanning in iOS, we let the OS to
+    // throttle the notifications of beacons found.
+    // scanner.stop()
```
wilsonpage: I think this needs a little more explanation.

We don't stop scanning when the app goes into the background as we want bluetooth scanning to continue. If we stop scanning we have no way of starting again until the app comes to the foreground. When the app is in the background and the scanner is still running, the OS will periodically call the scan-callback with results.

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,41 @@ import Foundation
...
   }
   
   func onItemFound(item: Dictionary<String, AnyObject>) {
-    print("item found", item["url"]);
+    print("item found", item["url"])
     bridge.eventDispatcher()
-      .sendDeviceEventWithName("magnetscanner:itemfound", body: item);
+      .sendDeviceEventWithName("magnetscanner:itemfound", body: item)
+    
+    notifyUser(item["url"] as! String)
```
wilsonpage, line 31: What's the `as! String` bit for?

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,41 @@ import Foundation
...
+    if (toNotify.contains(url)) {
+      return;
+    }
+    toNotify.append(url)
+    if (notifyTimer != nil) {
+      notifyTimer.invalidate();
+    }
+    
+    // Run this timer in the main loop otherwise our notifications won't show up.
+    notifyTimer = NSTimer.init(timeInterval: 60 ,
```
wilsonpage: Could you explain what `NSTimer` is and how we're using it

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,41 @@ import Foundation
...
+                               target: self,
+                               selector: #selector(MagnetScannerReact.doNotifyUser),
```
wilsonpage: Interesting syntax. What is `#selector`?

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,41 @@ import Foundation
...
+    
+    notifyUser(item["url"] as! String)
+  }
+  
+  // Throttles the notification process, waiting for 1 minute until
+  // setting up the badge with the number of elements nearby.
+  func notifyUser(url: String) {
+    if (toNotify.contains(url)) {
+      return;
+    }
```
wilsonpage, line 39: As you want to dedupe on url and order doesn't matter, would a `Dictionary` make more sense instead of an `Array`?

ios/ScannerBle.swift
```diff
@@ -25,7 +25,7 @@ import CoreBluetooth
 ///
 class BeaconScanner: NSObject, Scanner, CBCentralManagerDelegate {
   private var centralManager: CBCentralManager!
-  private let beaconOperationsQueue: dispatch_queue_t = dispatch_queue_create("beacon_operations_queue", nil)
+  private let beaconOperationsQueue: dispatch_queue_t = dispatch_queue_create("beacon_operations_queue", DISPATCH_QUEUE_SERIAL)
```
wilsonpage, line 30: Might be good to explain this with a comment as it caused you so much confusion.

ios/magnet/AppDelegate.m
```diff
@@ -11,6 +11,8 @@
 
 #import "RCTRootView.h"
 
+#import "magnet-Swift.h"
```
wilsonpage, line 15: What is this file? I don't see it in the patch.

ios/magnet/AppDelegate.m
```diff
@@ -52,8 +54,17 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
   [self.window makeKeyAndVisible];
   return YES;
 }
 
+- (void)applicationDidBecomeActive:(UIApplication *)application {
+  [NotificationsHelper clearNotifications];
+}
+
+- (void)applicationDidEnterBackground:(UIApplication *)application {
+  [NotificationsHelper clearNotifications];
```
wilsonpage, line 69: Is there a way to know if the app is in the foreground so we can prevent the notifications being dispatched?

ios/MagnetScannerReact.swift
wilsonpage: NIT: this kind of argument indentation stresses me out. Leads to visual inconsistency.

ios/MagnetScannerReact.swift
arcturus: Will copy & paste :P

ios/MagnetScannerReact.swift
arcturus, line 31: The dictionary is defined as `Dictionary<String, AnyObject>` but the function notify user needs a string. Basically we are casting `AnyObject` to string, as we know we have a string.

ios/MagnetScannerReact.swift
arcturus: is a way of specifying a callback, before swift 2.2 those were strings and prone to commit errors, now they are not, you can check:

https://developer.apple.com/library/ios/documentation/General/Conceptual/DevPedia-CocoaCore/Selector.html

ios/MagnetScannerReact.swift
arcturus, line 39: Actually no, a Dictionary is an structure that we don't need here, as it's a <key,value>, and we don't need to store anything as value.
Also  the arrays in swift have function like contains that is enough to tell us if we already have an element on it.

ios/ScannerBle.swift
arcturus, line 30: Yes, totally forgot and caused me so much problems until I figured it out.

ios/magnet/AppDelegate.m
arcturus, line 15: Yes it's a bit tricky, I'll comment on it on the code too.

As we are using a class in swift in the app delegate (the notifications), in order to include swift code into objective-c one you need that 'automagically' generated header:
<projectname>-Swift.h

ios/magnet/AppDelegate.m
arcturus, line 69: Good point, I though about it, but considered that is easier to clean on foreground and background, we will never get to an inconsistent state, but can dive into it.

arcturus, line 69: I don't want to mess with the timer stuff, as it can get tricky, but in the didEnterBackground we could do something like:
`[NotificationHelper disableNotifications]'

and then in the didBecomeActive:
`[NotificationHelper enableNotifications]'

What do you think @wilsonpage ?

wilsonpage, line 69: I don't mind really. I think eventually, once we know how all the parts work, we should create a ReactNative JS abstraction module around both android and iOS and control notifications from there.

``` js
Notifications.enable();
Notifications.disable();
Notifications.clear();
```

arcturus:
@wilsonpage r?

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,46 @@ import Foundation
   }
   
   @objc func stop() {
-    scanner.stop()
+    //    We don't stop scanning when the app goes into the background as we want bluetooth scanning to continue.
+    //    If we stop scanning we have no way of starting again until the app comes to the foreground.
+    //    When the app is in the background and the scanner is still running,
+    //    the OS will periodically call the scan-callback with results.
+    // scanner.stop()
```
wilsonpage: This line probably not necessary.

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,46 @@ import Foundation
   }
   
   @objc func stop() {
-    scanner.stop()
+    //    We don't stop scanning when the app goes into the background as we want bluetooth scanning to continue.
+    //    If we stop scanning we have no way of starting again until the app comes to the foreground.
+    //    When the app is in the background and the scanner is still running,
+    //    the OS will periodically call the scan-callback with results.
```
wilsonpage: Nit: strange intedentation

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,46 @@ import Foundation
...
+    if (notifyTimer != nil) {
+      notifyTimer.invalidate();
+    }
+    
+    // We need to throttle the notifications, waiting for receive more nearby web pages,
+    // so we use NSTimer to wait for 1 minute and after that we trigger the action, in
+    // this case doNotifyUser, that notifies depending on the number of web pages found
+    // during the waiting period.
+    notifyTimer = NSTimer.init(timeInterval: 60 , target: self, selector: #selector(MagnetScannerReact.doNotifyUser),
+                               userInfo: nil, repeats: false)
```
wilsonpage: NIT:

``` swift
notifyTimer = NSTimer.init(
  timeInterval: 60 ,
  target: self,
  selector: #selector(MagnetScannerReact.doNotifyUser),
  userInfo: nil,
  repeats: false)
```

ios/magnet/AppDelegate.m
```diff
@@ -52,8 +55,19 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
   return YES;
 }
 
+- (void)applicationDidBecomeActive:(UIApplication *)application {
+  [NotificationsHelper clearNotifications];
+  [NotificationsHelper disable];
+}
+
+- (void)applicationDidEnterBackground:(UIApplication *)application {
+  [NotificationsHelper clearNotifications];
```
wilsonpage, line 69: If our code is solid we shouldn't need this line?

ios/magnet/AppDelegate.m
```diff
@@ -11,6 +11,9 @@
 
 #import "RCTRootView.h"
 
+// Include the project headers for swift code. (<project-name>-Swift.h)
+#import "magnet-Swift.h"
```
wilsonpage, line 15: I'm still confused what this is for :) Where is `magnet-Swift.h` and why is it needed?

ios/MagnetScannerReact.swift
arcturus: I hate you Wilson Scott Page, I hate you.

ios/magnet/AppDelegate.m
arcturus, line 69: Trust me, with timers in the middle, better to be sure, specially cause what we are executing is not really resource consuming.

ios/magnet/AppDelegate.m
arcturus, line 15: Because AppDelegate.m is an objective-c file, in order to access swift code (NotificationsHelper.swift) the way of including the header is like that, <project-name>-Swift.h . Yes it's crappy but is the way apple provides:

https://developer.apple.com/library/ios/documentation/Swift/Conceptual/BuildingCocoaApps/MixandMatch.html

ios/MagnetScannerReact.swift
```diff
@@ -14,12 +17,49 @@ import Foundation
...
+    // We need to throttle the notifications, waiting for receive more nearby web pages,
+    // so we use NSTimer to wait for 1 minute and after that we trigger the action, in
+    // this case doNotifyUser, that notifies depending on the number of web pages found
+    // during the waiting period.
+    notifyTimer = NSTimer.init(
+      timeInterval: 60 ,
+      target: self,
+      selector: #selector(MagnetScannerReact.doNotifyUser),
+      userInfo: nil,
+      repeats: false)
```
wilsonpage, line 54: Ah, bliss 😄

wilsonpage:
R+

Sorry for all the questions 👅

PR #188, opened 2016-07-14
Started recording URL detections in local History DB (closes #184)

### LINKED ISSUE #184: Store found URLs in a local 'history' database
If we want to make notifications more intelligent, we need to have a way of knowing what was found previously so as not to annoy the user, notifying them of stuff they already know about.

wilsonpage:


wilsonpage:
@arcturus r?

android/app/src/main/java/org/mozilla/magnet/db/History.java
```diff
@@ -0,0 +1,192 @@
...
+
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
+
+import static android.provider.BaseColumns._ID;
+import static org.mozilla.magnet.db.Schema.HistoryEntry.COLUMN_NAME_TIME_LAST_SEEN;
+import static org.mozilla.magnet.db.Schema.HistoryEntry.COLUMN_NAME_URL;
+import static org.mozilla.magnet.db.Schema.HistoryEntry.COLUMN_NAME_TIME_FIRST_SEEN;
+
+public class History extends SQLiteOpenHelper {
```
arcturus: Can we change the approach and make History not coupled with the SQLiteOpenHelper?

My point is the History class will have some methods like record, update, updateLastSeen, that has some semantic meanining. That should not be tied to the use of SQLite, we could pass a dependency in the constructor of History, that does the actual data saving and processing. That will help us to change the implementation for any other if needed and better unit tests when we have them :)

For example:

``` java
public class History {
    public History(Context context, DataModel model) {}
}

public interface DataModel {
    public void onCreate() {}
    public void onUpgrade() {}
    public void insert() {}
....
}

public class SQLiteDataModel implements DataModel {
...
}
```

Also we could have a factory to get the History objects already initialized with the strategy we want:

``` java
public class HistoryFactory {
     public static History getHistory(Context ctx) {
        return new History(ctx, new SQLiteDataModel());
        // Or even better a singleton.
     }
}
```

wilsonpage: I'm not sure I follow. I used [this tutorial](http://mobilesiri.com/android-sqlite-database-tutorial-using-android-studio/) to guide me .

arcturus: The sql part looks ok, but my point is that History should not do any sql operations, but delegate them to a class that performs them.

That class should be a dependency that we inject when we create the History object.

Does it make sense?

wilsonpage: OK, so instead we pass in a subclass of `SQLiteOpenHelper` (say `HistoryDB`) to `History` that includes some high level methods like: `find()`, `save()`, `update()`?

arcturus: Yeah, I would call it HistorySQLite, so if we realise that we need to replace SQLite by ... XYZ, we just replace the HistoryXYZ object that implements the logic.

But we keep the high level semantics of History, the want that we want to be used by other classes.

android/app/src/main/java/org/mozilla/magnet/database/History.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.Date;
+import java.util.concurrent.TimeUnit;
+
+public class History {
+    private static final String TAG = History.class.getName();
+    private static final long RECENT_TIME_PERIOD = TimeUnit.HOURS.toMillis(1);
+    private HistoryDatabase mDB;
+
+    /**
+     * Factory to create new `History` interface.
```
arcturus: nit: Factory to create new `History` object.

android/app/src/main/java/org/mozilla/magnet/database/History.java
```diff
@@ -0,0 +1,75 @@
...
+     *
+     * @param context
+     * @return History
+     */
+    public static History create(Context context) {
+        return new History(new HistoryDatabaseSQL(context));
```
arcturus: We could do this a singleton, but don't see any other objects using this class so far, so we can leave it.

android/app/src/main/java/org/mozilla/magnet/database/History.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    /**
+     * Create a new `History` interface passing
```
arcturus: nit: object not interface

android/app/src/main/java/org/mozilla/magnet/database/History.java
```diff
@@ -0,0 +1,75 @@
...
+     */
+    public void record(String url) {
+        Log.d(TAG, "add: " + url);
+        HistoryRecord recentRecord = getRecent(url);
+
+        // when a recent record has been found just
+        // update it instead of creating another
+        if (recentRecord != null) {
+            Log.d(TAG, "found recent record");
+            mDB.updateLastSeen(recentRecord.id);
```
arcturus: Should we just pass the record object to `updateLastSeen` ?

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabase.java
```diff
@@ -0,0 +1,9 @@
+package org.mozilla.magnet.database;
+
+import java.util.Date;
+
+interface HistoryDatabase {
```
arcturus: nit: public interface ...

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabaseSQL.java
```diff
@@ -0,0 +1,158 @@
...
+        SQLiteDatabase db = this.getWritableDatabase();
+        long now = System.currentTimeMillis() / 1000;
+
+        ContentValues values = new ContentValues();
+        values.put(COLUMN_NAME_URL, url);
+        values.put(HistoryEntry.COLUMN_NAME_TIME_FIRST_SEEN, now);
+        values.put(COLUMN_NAME_TIME_LAST_SEEN, now);
+
+        db.insert(HistoryEntry.TABLE_NAME, null, values);
+        db.close();
```
arcturus: Do we need to close? Or should we reuse the db object? Not familiar with latest sqlite implementation.

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabaseSQL.java
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    /**
+     * Return a HistoryRecord from the given cursor.
+     *
+     * @param cursor
+     * @return HistoryRecord
+     */
+    private HistoryRecord toHistoryRecord(Cursor cursor) {
+        int id = cursor.getInt(cursor.getColumnIndex(_ID));
```
arcturus: Should this function return a List<HistoryRecord> so if we have a cursor with more than 1 element we return the list of HistoryRecord.

Not used now, but in the future could be interesting.

arcturus:
I think this first implementation is perfect just what we need. With the change for abstracting the storage solution we could swap in a future to use Content Providers (https://developer.android.com/guide/topics/providers/content-providers.html), in the case that the information we store is useful for any other app.

But as I said, so far looking pretty good.

@wilsonpage  left some comments, some of them not mandatory to implement.

Great work mate!

android/app/src/main/java/org/mozilla/magnet/database/History.java
wilsonpage: Any reason to pass more data to a function than it needs?

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabase.java
wilsonpage: Oh, how come it works without `public` keyword?

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabaseSQL.java
wilsonpage: Interesting point. I'm really not sure. If we did keep the reference open then I think we would need to re-architect slightly as when we have more than one table, they would need to share this reference. I'm guessing we'd also need to `.close()` the connection when the app is destroyed?

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabaseSQL.java
wilsonpage: If it's needed in the future yes. Instead of changing this, I would create a second function that calls `toHistoryRecord()` for each item in the list.

android/app/src/main/java/org/mozilla/magnet/database/History.java
arcturus: Cause to me, you update a record, so makes sense to pass the record. Also later on, for any additional logic, we could need more attributes from that object to perform the update ... imagine, restrictions or something similar.

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabase.java
arcturus: It does work cause it has default visibility. I don't know if you were intended to do that, but that interface will be available just to the classes that are in that package: http://docs.oracle.com/javase/tutorial/java/javaOO/accesscontrol.html

If that's your intention, then leave as it is :)

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabaseSQL.java
arcturus: Fair enough

android/app/src/main/java/org/mozilla/magnet/database/HistoryDatabase.java
wilsonpage: Oh yeah, I knew that 😉

wilsonpage:
@arcturus another iteration to try to keep a single static database open. [This article](http://touchlabblog.tumblr.com/post/24474750219/single-sqlite-connection) suggests that it's fine to keep the DB open until the app or service closes.

wilsonpage:
It doesn't feel quite right yet. I think the thing that's confusing is two way relationship between the `HistoryTable` and `DatabaseSQL`. `HIstoryTable` depends on `DatabaseSQL` to get a handle on the database, and `DatabaseSQL` depends on `HistoryTable` so that it knows the `Schema` to create the table with.

wilsonpage:
Perhaps during init, before the table is created, the `History` class should pass/inject the table structure into `DatabaseSQL` so that it can be created in the `onCreate` handler. That way we'd maintain a sensible dependency graph.

wilsonpage:
OK I've swapped a few things around and now we don't have cyclic dependencies any more. I think we're ready to land.

android/app/src/main/java/org/mozilla/magnet/database/History.java
```diff
@@ -0,0 +1,81 @@
...
+     * We use a factory here instead of calling the
+     * constructor directly to allow us to pass
+     * a mock `HistoryStore` object into the
+     * constructor for testing.
+     *
+     * @param context
+     * @return History
+     */
+    public static History get(Context context) {
+        HistoryStore store = DatabaseSQL.get(context).getHistoryTable();
```
arcturus, line 28: Can we have a singleton here?

arcturus:
Shiiiiiiip it!

r+!

PR #195, opened 2016-07-19
Add distance calculation to match android (closes #177)

### LINKED ISSUE #177: Calculate distance in magnet-scanner-ios
If we are to show live distances in the app UI we need the ios scanner library to calculate and expose these to ReactNative JS land, like the android equivalent library is doing.

wilsonpage:


wilsonpage:
@arcturus r?

I've used that same calculation as we have in Android. Looking across the web it seems there is still a lot of confusing regarding the best way to calculate distance. In [eddystone-ios](https://github.com/BlueBiteLLC/eddystone-ios/blob/78b38fe804b94db87bfc6cd926bbfa2aecc349d6/Pod/Classes/Beacon.swift#L76-L88) their calculation returns a very high number that couldn't possibly be meters.

arcturus:
Works for me.

Much clearer now, shiiiiiiiip it!

r+!

Issue #189, opened 2016-07-15
Notifications - Opening app from a notification should open it in list view


isabelrios:
As commented by email..
It could happen that you open a tile in expanded view. Then leave the app in background. The notification appears on the screen and tapping on it takes you to the app to the expanded tile, not to the list view.

oteo:
@isabelrios once we removed the expanded feature in the client, wouldn't this issue be removed?

isabelrios:
Yes, we can close this if we are not going to have the expanded view anymore.

isabelrios:
Once we have again the POI expanded view, this issue is happening.
Steps:
Fake the position to a place with several POIs
Wait to receive the notifications
Tap on one of the notifications and open the POI expanded view
Go back to utility tray and tap on other notification about a different POI

Expected:
The app is open showing that specific POI or the general list view to see all POI

Actual:
The app is open in the first expanded POI.

isabelrios:
There is something similar when you left the app in background in the subscriptions view. 
When you have a notification and tap on it to open the app, it is not open on the POI the notification is about or in the main list view, the app is open in the subscriptions view

isabelrios:
This will be solved at least for android once issue #331 lands

oteo:
The issue in Android is solved, after landing #331 but the issue is still happening in iOS

oteo:
The issue continue happening in iOS after landing PR #355 but I am going to open a new one to report it

PR #198, opened 2016-07-29
Simplified MVP with advanced features preffed off

wilsonpage:
Added flags


wilsonpage:
@isabelrios @arcturus this is the greatly simplified MVP based on the simplified 'ux-v2' list design. Hoping we can ship this is in the PlayStore ASAP and gradually pref on features when they're ready. Perhaps we could even A/B test them somehow?

Please build and review :)

wilsonpage:
Ah I just remembered that the About/Settings menu is not really finished. I'm not sure what the exact requirements are, but I think we may need to link to Terms and Privacy policies. Does this have to be embedded in the app or can we link out?

isabelrios:
I think there could be a link to Terms and Privacy policies to external sites, not embedded in the app

isabelrios:
@wilsonpage the new tiles look really nice! :)

I have seen a few things in android though with this new version of the app:
- It takes a while for the tiles to be shown, they change from a default form to the final one, please see this video: https://youtu.be/Ln7LeuXaraw 
  Not sure if that time could be reduced or directly show the tiles when completely formed
- When pulling down to refresh it inmediately appears 'Nothing found' and then 'Scanning', would it be possible to remove the 'Nothing found' there?
- I know you said the About/Setting menu is not finished, but just in case :). Although there is the close button, seems that all that header where the button is is active as tapping wherever you want takes back to the main view
- The view source tile is shown fine but as you can see in the video the letters are overlapped with the content of the tile. Not sure if there is something to be done there but just to be aware that it could happen with some tiles and the view is not as nice as the other

Will continue testing this and the iOS version. Thanks!

lib/views/list-item/index.js
```diff
@@ -272,7 +263,7 @@ class ListItem extends React.Component {
       : clamp(vx * -1, min, max) * -1;
 
     debug('swiping away ...', vx, velocity);
-    Metrics.swiped(this.props.unadaptedUrl);
+    Metrics.swiped(this.props.item.unadaptedUrl);
```
isabelrios, line 266: I guess this metric is not necessary now

isabelrios:
Hi @wilsonpage more feedback after checking the iOS app:
- The transition to open and close the setting menu sometimes is not very soft, here what I can see: https://youtu.be/jffbkp1Z55k
- Same as in android, although there is a close button in the Setting view, all the header is active and tapping on it takes back to main view
- Also same behaviour than in android when pulling to refresh with the 'Nothing found' string

isabelrios:
A few more things:
- Similarly to issue #189, when opening the app from a notification if the setting view was open that is the view shown, no the main view. This would not be a very common scenario but just to be aware of it
- Sometimes the setting or close button is not responding. I got to reproduce it by opening the app, then opening the settings view and sending the app to background, after some minutes I brought the app back to foreground and tried to tap on the close and then on setting button

lib/views/list-item/index.js
wilsonpage, line 266: Swipable list-items is still in the code base but preffed off.

isabelrios:
Settings Menu looks better and the issue to close only with the dedicated button is solved. Also the transition in iOS is working much better.
It is also fixed the issue when pulling down to refresh, now it shows Scanning directly.

I have one doubt about android, when being in Settings Menu if you tap on the go back key, it sends the app to background and when bringing it back to foreground it starts Scanning in main view. Not sure if that is the expected behaviour there. For me is more intuitive if the go back button takes back from settings to main view, but it is just an opinion :)

Great work @wilsonpage! :)

config.js
```diff
@@ -1,4 +1,10 @@
 module.exports = {
+  "flags": {
+    "injectTestUrls": true,
```
isabelrios: We have to remember to turn this back to false before having the final version.
Or change the default urls in case we want to add a default one explaining the app.

lib/action-creators.js
```diff
@@ -0,0 +1,49 @@
+
```
arcturus, line 1: Sorry for being a bit picky, can we have a tiny bit of dock describing what this file does?

wilsonpage, line 1: Sure. As part of this patch (before we started MMMVP) I moved all our state into a single Redux 'store'. This means that we have no (or little) state living in the rest of our views or modules. Redux stores allow you to only change the application state via defined 'actions'. This file acts as an interface to the store that other modules can use to change app state.

Redux has been embraced by the React community as the best way to manage state. So much so that Facebook hired the author of Redux.

ios/MagnetScannerReact.swift
```diff
@@ -17,10 +17,7 @@ import Foundation
   }
   
   @objc func stop() {
-    // We don't stop scanning when the app goes into the background as we want bluetooth scanning to continue.
-    // If we stop scanning we have no way of starting again until the app comes to the foreground.
-    // When the app is in the background and the scanner is still running,
-    // the OS will periodically call the scan-callback with results.
+    scanner.stop()
```
arcturus, line 20: After reading the rest of the PR, saw that we moved the stop to the app logic.
I think this is the correct Idea, as we keep this interface as agnostic as possible.

Good call.

lib/scanner/index.js
```diff
@@ -81,30 +77,22 @@ class Scanner extends Emitter {
    * @private
    */
   onItemFound(item) {
-    var url = item.url;
+    var shouldPopulate = this.shouldPopulateItem(item.url);
```
arcturus, line 80: This can be null, if the option objects doesnt contain the `shouldPopulateItem` this may fail.

lib/store.js
```diff
@@ -0,0 +1,98 @@
+'use strict';
+
```
arcturus, line 2: More documentation in general, I know you left some comments explaining it on gh comments, but for possible contributors to read will be interesting.

lib/scanner/index.js
wilsonpage, line 80: Is there value in this if we're currently the only consumer of this module? It's hard to know when to stop. If this we're a standalone library used by other projects, or even just used in more than one place in our project, I'd understand.

lib/views/header-bar.js
```diff
@@ -8,37 +8,35 @@ var ReactNative = require('react-native');
...
+            source={require('../images/header-logo.png')}/>
         </TouchableOpacity>
         {this.renderRight()}
       </View>
     )
   }
 
   componentWillUpdate() {
-    LayoutAnimation.easeInEaseOut();
+    // LayoutAnimation.easeInEaseOut();
```
arcturus, line 35: Dead code?

lib/views/header-bar.js
```diff
@@ -76,38 +74,32 @@ class HeaderBarView extends React.Component {
 
   renderRight() {
     return <View style={styles.right}>
+      <TouchableOpacity
```
arcturus, line 77: Is this causing the perf problems that @isabelrios was showing in the video?

wilsonpage, line 77: Perf problem now fixed. It was related to HTTP requests blocking JS thread during animation.

lib/views/list-item/index.js
```diff
@@ -207,9 +227,7 @@ class ListItem extends React.Component {
     debug('gesture end', vx, exceededThreshold);
 
     // fast swipes fling the tile away
-    if (exceededThreshold) {
-      this.swipeAway(vx);
-    }
+    if (exceededThreshold) this.swipeAway(vx);
```
arcturus, line 230: Hate it! 1 liners with ifs ... I hate them :P

lib/views/list-item/index.js
```diff
@@ -179,8 +200,7 @@ class ListItem extends React.Component {
 
   onMoveShouldSetPanResponder(e, { dx, dy }) {
     debug('panresponder move should set?', dx, dy);
-    if (this.props.expanded) return false;
-
+    if (!this.props.swipable) return false;
```
arcturus, line 203: \o/ pref-ing features for the win!

lib/views/list-item/index.js
wilsonpage, line 230: Sowweee 😉  That's the current style in this project, if we want to change it, it should be changed everywhere. I could compromise with `if (foo) { return bar; }`, but using three lines when you could use one seems like  unnecessary clutter.

We're lean now, so stuff like this shouldn't matter at this stage 😛

arcturus, line 230: That's why i'm not blocking this pr ... i was just expressing my angriness ... you own me a ☕

isabelrios:
Thanks for the changes!The back button from settings takes you to the app which seems more coherent and the notifications also look nice with the new strings.

arcturus:
Pretty good work, did some testing in ios and android, and tbh, this is what we should have done from the very beginning!

🚢 🚢 🚢 🚢 :shipit:

PR #202, opened 2016-08-04
Linking elements in the settings page with external urls r=wilsonpage

arcturus:
@wilsonpage added the links in the config file to make it easier to change ... if we have to ;)

### DIFF
--- a/lib/views/menu-scene.js
+++ b/lib/views/menu-scene.js
@@ -6,7 +6,7 @@
 const debug = require('../debug')('MenuScene');
 const ReactNative = require('react-native');
 const { connect } = require('react-redux');
-const { theme } = require('../../config');
+const { theme, settings } = require('../../config');
 const React = require('react');
 
 const {
@@ -15,6 +15,7 @@ const {
   Image,
   View,
   Text,
+  Linking
 } = ReactNative;
 
 class ListScene extends React.Component {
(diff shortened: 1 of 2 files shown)


lib/views/menu-scene.js
```diff
@@ -69,6 +78,16 @@ class ListScene extends React.Component {
   onClosePress() {
     this.props.navigator.pop();
   }
+  onItemPress(id) {
+    const url = settings.links[id];
+    if (!url) {
+      return;
+    }
```
wilsonpage: _cough_

lib/views/menu-scene.js
```diff
@@ -69,6 +78,16 @@ class ListScene extends React.Component {
...
+
+    Linking.openURL(url).catch(err => {
+      console.info(`Could not open url ${url} `, err);
+    });
```
wilsonpage, line 88: Under what circumstance would this fail? I imagine only for malformed URLs. In which case, worth it?

wilsonpage:
Arrrr pluuussssss ;)

lib/views/menu-scene.js
```diff
@@ -69,6 +78,16 @@ class ListScene extends React.Component {
   onClosePress() {
     this.props.navigator.pop();
   }
```
wilsonpage, line 80: NIT: line-break between blocks.

wilsonpage:
After nits :)

lib/views/menu-scene.js
arcturus: I did it in purpose, can i keep it? It's so nice and stylish

arcturus:
Ok, Ok, it's really hard for me to write 

``` javascript
if (!url) { return; }
```

PR #205, opened 2016-08-05
Analytics all the things

wilsonpage:


wilsonpage:
@isabelrios r?

lib/app.js
```diff
@@ -42,12 +43,7 @@ class AppView extends React.Component {
   constructor(props) {
     super(props);
 
-    // Metrics
-    Metrics.launch();
```
isabelrios, line 46: If possible I would  not remove the metrics, it would be better first check that we have enough data from ga and in that case, if these metrics do not give us anything else we need, we can remove them

wilsonpage, line 46: OK I can add that event in until we are confident that we can get what we need from GA 'session' data.

lib/app.js
```diff
@@ -42,12 +43,7 @@ class AppView extends React.Component {
...
-    Metrics.install();
```
isabelrios, line 47: Same as above please

build-scripts/android/build.sh
```diff
@@ -2,11 +2,11 @@
 
 set -e
 cd android
-./gradlew check
+./gradlew app:check
 
 if [[ $BUILD_RELEASE == 1 ]]; then
     ./gradlew assembleRelease --info --console plain | tee
 else
-    ./gradlew build --info --console plain | tee
```
isabelrios, line 10: I see travis is now green, are we skipping all the lint checks with this? I am afraid that we have removed many tests and if we skip also the lint we will not have many alerts when landing new code but we could live with this for landing this PR and work on the tests and the lint issue later on in a follow up

lib/notifications.js
```diff
@@ -0,0 +1,58 @@
+'use strict';
```
isabelrios, line 1: Sorry I am not sure I understand what this notifications.js does

build-scripts/android/build.sh
isabelrios, line 10: I have been testing and the metrics are recorded successfully both in android and iOS \o/
I only have a few questions and would prefer to have the other metrics in the code till we compare both and agree on remove them. 
That would be all and this could land :)

wilsonpage, line 10: We are still linting our codebase, just not our dependencies. `./gradlew app:check` performs linting. The ./gradlew build` was also performing the same lint checks (across our app and all dependencies) so I decided to skip.

I think we should submit a PR to the problematic dependency and if/when that lands we could begin linting deps again.

lib/app.js
```diff
@@ -42,12 +43,8 @@ class AppView extends React.Component {
...
-
+    track.appLaunch();
```
wilsonpage, line 46: @isabelrios here is where we're tracking 'app-launch'. Inside this function I also track 'first-app-launch'. I renamed this from 'install' to make it clearer what was actually being tracked; although you could assume that this is a similar time that the install completed.

isabelrios:
Checked that the events are recorded with the final version, both in android and iOS, seems to be recording all what was implemented :)
Lets land it! r+
Thanks! @wilsonpage

PR #206, opened 2016-08-08
adding project

isabelrios:


isabelrios:
Hi @wilsonpage if IRC we have to add 'Project' before magnet where it appears so I changed these two. It is still pending the logo in main view, I am not sure how to change that but if you guide me I can try.
r? thanks

wilsonpage:
Currently the logo is a PNG` so requires export out of a graphics source file I have. We could try replacing the png with a real font, like on the settings panel.

isabelrios:
I will try that option to change the name, thanks!

PR #209, opened 2016-08-10
Update to support latest metadata-service

wilsonpage:


wilsonpage:
Landed as 764d0f5ae61a4fb984be9d7e07b568d522e619f6

PR #212, opened 2016-08-16
Make Magnet for iOS build on case sensitive file systems. r=francisco

ferjm:
I am getting this XCode build error on a case sensitive file system:

clang: error: no such file or directory: '/Volumes/mozilladev/magnet/magnet-client/ios/Magnet/main.m'
clang: error: no input files

The fix is simple: `s/magnet*/Magnet*/g`

I tested this patch in a case insensitive file system and it seems to work fine.


arcturus:
r+!

Shipping it!
:shipit: :shipit:  🚢

PR #213, opened 2016-08-16
Use magnet-scanner-ios. r=francisco

ferjm:


arcturus:
Tested and working perfectly!

Thanks for the contribution, :shipit: !

PR #214, opened 2016-08-18
Logging notification clicks in ios r=wilsonpage

arcturus:
Hei @wilsonpage I enabled the tracking of the clicking in the notifications for ios.

This will be the first step to make the navigation go back to the first page when the notification is clicked.

### DIFF
--- a/ios/Magnet/AppDelegate.m
+++ b/ios/Magnet/AppDelegate.m
@@ -11,6 +11,8 @@
 
 #import "RCTRootView.h"
 
+#import "RCTBridge.h"
+
 // Include the project headers for swift code. (<project-name>-Swift.h)
 #import "magnet-Swift.h"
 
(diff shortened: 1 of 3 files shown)


ios/Magnet/AppDelegate.m
```diff
@@ -57,6 +59,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   self.window.rootViewController = rootViewController;
   [NotificationsHelper register];
   [self.window makeKeyAndVisible];
+  self.bridge = [[RCTBridge alloc] initWithBundleURL:jsCodeLocation
+                                            moduleProvider: nil
+                                            launchOptions:launchOptions];
```
wilsonpage, line 64: Can you explain what this line does?

wilsonpage:
r+ after discussed changes   --  🚗 it!!!

PR #215, opened 2016-08-18
Add distance sorting and item expiration behind a user-flag

wilsonpage:


wilsonpage:
@arcturus f/r?

Let me know if you have anything missing you want me to document. Again, apologies for the large patch, but I don't like hacking big features into an app :)

wilsonpage:
Haven't had chance to test on ios, will do tomorrow morn.

config.js
```diff
@@ -2,7 +2,24 @@ module.exports = {
...
     "injectTestUrls": false,
     "itemsExpandable": false,
-    "itemsSwipable": false
+    "itemsSwipable": false,
+  },
+
+  "userFlags": {
+    "sortByDistance": {
+      value: false, // default
+      title: 'Sort by distance',
```
arcturus, line 11: So far as we are not thinking on l10n the app is ok, but when we do it we will need to revisit this.

Anyway, so far for the experiment is perfect.

config.js
```diff
@@ -13,6 +30,9 @@ module.exports = {
     }
   },
 
+  "itemExpires": 60000, // 1 min
+  "itemExpiring": 10000, // 10 secs
```
arcturus, line 34: What's the difference between expires and expiring?

lib/app.js
```diff
@@ -35,47 +40,112 @@ const {
...
+    this.scanner = new Scanner({
+      onUpdate: this.props.updateItem,
+      onLost: this.props.removeItem,
+      onMetadata: this.props.setItemMetadata,
+      expiryEnabled: this.enabled.bind(this, 'removeOldItems'),
+      getItems: () => this.props.items
     });
 
     this.setupNotificationListeners();
+    this.inflateUserFlags();
```
arcturus, line 74: Seeing that we import everything and manipulate the settings view here in the main container, I have the feeling that we load absolutely the whole ui.

Fortunately we don't have a heavy ui so hope this wont be a problem, but in the future,  once we move towards more product and less experiment ;P, we can think on load view dynamically ;)

wilsonpage, line 74: Don't follow. Where is the SettingsScene being manipulated?

config.js
wilsonpage, line 34: - 'expiring' === 'dying'
- 'expired' === 'dead'
- 'expiring' tiles are greyed out
- 'expired' tiles are removed completely

lib/app.js
```diff
@@ -160,29 +273,142 @@ class AppView extends React.Component {
...
+  }
+
+  /**
+   * Called when a list-item is pressed.
+   *
+   * Depending on what prefs are set, this
+   * could open the expanded view, or simply
+   * navigate to the item's url.
+   */
+  onItemPress(item) {
```
arcturus, line 308: Have the general feeling that now app.js is a huge controller for all the views, a bit mixed for me, but won't block on this. We can revisit if we find a better way.

lib/scanner/index.js
```diff
@@ -43,9 +57,44 @@ class Scanner extends Emitter {
   stop() {
     debug('stop');
     MagnetScanner.stop();
+    this.stopExpireCheck();
     this.listener.remove();
   }
 
+  startExpireCheck() {
+    this.nextExpireCheck = setTimeout(() => {
```
arcturus, line 65: Should we do the `if (!this.expiryEnabled()) return;` here so we avoid adding the setTimeout and the execution of code that later on is going to return?

arcturus, line 65: Humm, I'm starting to think that you leaved like that to be able to enable/disable without having to do a `startExpireCheck` again.

lib/scanner/index.js
```diff
@@ -43,9 +57,44 @@ class Scanner extends Emitter {
...
+  stopExpireCheck() {
+    clearTimeout(this.nextExpireCheck);
+  }
+
+  checkExpired() {
+    if (!this.expiryEnabled()) return;
+    
+    var now = Date.now();
+    this.getItems().forEach(item => {
+      var isMdns = item.distance === -1;
```
arcturus, line 80: nit: `const isMdns ...`

lib/scanner/index.js
```diff
@@ -43,9 +57,44 @@ class Scanner extends Emitter {
...
+
+      // when an item expires,
+      // it is removed from the list
+      if (age > config.itemExpires) {
+        debug('item expired', item.id);
+        return this.onLost(item.id);
+      }
+
+      // when an item is 'expiring' it is greyed out
+      var expiring = age > config.itemExpiring;
```
arcturus, line 93: nit: const

lib/app.js
arcturus, line 74: Not manipulated, but loaded in memory, even if we don't access settings.

Again, nothing to worry right now, but something to think in the future.

lib/scanner/index.js
```diff
@@ -108,6 +159,10 @@ class Scanner extends Emitter {
         this.emit('networkerror', err);
       });
   }
+
+  itemIsNew(id) {
+    return !this.getItems().some(item => item.id === id);
```
arcturus, line 164: Don't know if it's better to use some or find here.

lib/app.js
wilsonpage, line 308: An item press can result in an app navigation (when expanded view is preffed on), so we do have to bubble this event up the the app level sometimes. Perhaps though we could handle logic at the list view level and only bubble up the app view level when need be.

That said (I'm thinking out loud :)) we could use the passed `navigator` object into the `ListScene` and let it do the navigation. Or we also may be able to do routing via a rdux state change.

So many options!!

lib/scanner/index.js
wilsonpage, line 65: Yeah, we could do the check earlier, but we'd need a hook to kick the scanner to check for expiries again when the user changes that option. That's all possible though :)

lib/app.js
wilsonpage, line 74: Although we don't render the settings menu, we need the `userFlags` loaded from storage to make decision about how the app should run.

Example:
- User turns on `sortByDistance`.
- CLoses the app
- Launches the app
- They would expect this pref to be persisted.
- We need to load their prefs to know that we should sort by distance

arcturus, line 74: Sure, I agree with loading the user flags, I was talking about loading in memory the Settings Scene.

wilsonpage, line 74: I'm not following. What line is that happening? This line is only related to loading `userFlags`.

wilsonpage, line 74: The `SettingsScene` should only be rendered when the `'settings'` route is hit.

arcturus, line 74: Yes, but my point is that is loaded in memory even before we need to render it. Again, nothing we should care about optimising now, but think about the future.

arcturus:
r+ trying it on ios and android and looking really nice.

Also the removing from the list works pretty well, personally would reduce the time from expiring to expired a bit, but so far looking good, maybe Tif could comment on that.

:shipit:

arcturus:
@wilsonpage there is a test failing, can you take a look?

PR #228, opened 2016-09-02
Fix bluetooth prompt 'deny' loop (closes #225)

### LINKED ISSUE #225: Bluetooth prompt is very agressive
Launching the app with Bluetooth turned off prompts the user to turn it on. If they deny the prompt appears again, repeatedly. The only way to escape it is to close the app.

The prompt should not show again if the user denies the request to turn on Bluetooth.

wilsonpage:


lib/app.js
```diff
@@ -58,8 +58,6 @@ class App extends React.Component {
   constructor(props) {
     super(props);
 
-    this.scanning = false;
```
wilsonpage, line 61: State moved inside `Scanner`

PR #250, opened 2016-10-13
Fix notification logging.

### LINKED ISSUE #244: Event action 'app-launch-from-notification' is not recorded for iOS
This is happening in both dev and market version. The information about launching the app from the notification is not being recorded.

arcturus:
Closes #244

### DIFF
--- a/ios/Magnet/AppDelegate.m
+++ b/ios/Magnet/AppDelegate.m
@@ -54,10 +54,7 @@
 }
 
 -(void)application:(UIApplication *)application didReceiveLocalNotification:(UILocalNotification *)notification {
-  if (application.applicationState == UIApplicationStateInactive ||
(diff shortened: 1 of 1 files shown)


arcturus:
@wilsonpage r?

We had a stupid comparison, that's valid just when the app is not running, but we want to notify no matter what's app status.

wilsonpage:
Nice catch :)

Unrelated: Ideally we shouldn't be writing code in `AppDelegate.m` as this is really RN territory. Would be better if we can move all out code into separate modules. in Android we have to much code in `MainActivity.java`, that too needs moving. Not urgent, just a thought.

wilsonpage:
Shiiiiip it +r

arcturus:
Definitely a good practice, but we will let the refactors for post-gate1? ;)

PR #259, opened 2016-10-18
Add geolocation scanning (closes #258)

### LINKED ISSUE #258: Add geolocation based scanning (android)

wilsonpage:


wilsonpage:
Depends on https://github.com/mozilla-magnet/magnet-scanner-android/pull/9

wilsonpage:
This is a significant change in scanning architecture:
- All background scanning logic is contained inside `magnet-scanner-android`
- Apps can still interfacing directly with the `MagnetScanner` class.
- Apps also have the option of calling `scanner.startBackgroundScanning()` and listening for a broadcast to wake up their services (such as background notification dispatch).
- We no longer scan in the background using Alarms, but instead rely on the logic inside magnet-scanner-android`to trigger our app's`ReceiverMagnetScannerItemsFound` receiver.
- `magnet-scanner-android` runs background scans based on a combination of interval and device distance changes. There are lots of knobs to tweak this logic. We need to stike a balance between accuracy, frequency and power consumption. This will take some trial and error.

wilsonpage:
@arcturus f?

PR #276, opened 2016-10-27
Update to allow http image loads in ios (fixes #274)

### LINKED ISSUE #274: Not shown image in the tile for atlasobscure source
I am not able to see the picture in atlasobscure tile for iOS.
The strange thing is that the picture is shown correctly for the tile in android
@arcturus can you have a look at it?

e.g. http://www.atlasobscura.com/places/the-churchill-war-rooms

**Tile in iOS**
ios_atlasobscure
**Tile in Android**
android_atlasobscure

wilsonpage:


wilsonpage:
@arcturus r?

ios/Magnet/Info.plist
```diff
@@ -26,21 +26,17 @@
 	<dict>
 		<key>NSAllowsArbitraryLoads</key>
 		<true/>
```
wilsonpage, line 28: Interestingly it appears as though `NSAllowsAbititraryLoads` was set to `true` all along. Although our info.plist was malformed, in that it had two `<key>NSAppTransportSecurity</key>` sections.

wilsonpage: In-fact `NSAllowsAbititraryLoads: true` is the  default for react-native:
https://github.com/facebook/react-native/blob/master/Examples/2048/2048/Info.plist#L25-L29

arcturus (approved):
r+ shiiip it

PR #282, opened 2016-10-28
Sending the correct applaunch event in ios

### LINKED ISSUE #244: Event action 'app-launch-from-notification' is not recorded for iOS

arcturus:
Fixes #244

### DIFF
--- a/ios/Magnet/AppDelegate.m
+++ b/ios/Magnet/AppDelegate.m
@@ -61,7 +61,7 @@
 }
 
 -(void)application:(UIApplication *)application didReceiveLocalNotification:(UILocalNotification *)notification {
-  [self.bridge.eventDispatcher sendDeviceEventWithName:@"notification:applaunch" body:nil];
+  [self.bridge.eventDispatcher sendDeviceEventWithName:@"applaunch" body:nil];
 }
 
 @end


arcturus:
@isabelrios could you take a look to check if we finally get this events in ios?

ios/Magnet/AppDelegate.m
```diff
@@ -61,7 +61,7 @@ - (void)applicationDidEnterBackground:(UIApplication *)application {
 }
 
 -(void)application:(UIApplication *)application didReceiveLocalNotification:(UILocalNotification *)notification {
-  [self.bridge.eventDispatcher sendDeviceEventWithName:@"notification:applaunch" body:nil];
+  [self.bridge.eventDispatcher sendDeviceEventWithName:@"applaunch" body:nil];
```
wilsonpage, line 64: I'd be surprised if that worked. Fro two reasons:
1. The event name is wrong https://github.com/mozilla-magnet/magnet-client/blob/master/lib/notifications.js#L33
2. The JS app is likely not running yet to receive the notification

I think you need to implement native method [`.appLaunchedFromNotificaiton()`](https://github.com/mozilla-magnet/magnet-client/blob/master/lib/notifications.js#L57) to allow the JS code to ask the native side if if was launched from a notification.

In android we have the event from when the app is 'brought back to the foreground' and the method for when the app is 'first launched' from a notification.

arcturus:
Thanks for the tip @wilsonpage will try to implement the suggestion.

isabelrios:
@arcturus as @wilsonpage thought that is still not working. Will check it when the new implementation is ready.

wilsonpage:
Closing this for now. Feel free to reopen if we're still planning to land :)

PR #285, opened 2016-10-31
Load from local bundle by default

### LINKED ISSUE #284: Magnet app cannot be open on iOS when WiFi has been disconnected
Finally with Maria's help we got steps to reproduce this issue that seems 100% reproducible.
Please see video attached:
https://youtu.be/X03zNyawO64

Pre-requisites: 
Magnet is not launched. It could have been launched before, but closed so it is not in background

Steps: 
1-Disable your wifi and enable it again
2-Launch Magnet

The only way to recover the app is:
- After following previous steps you can see the app in background, kill it
-Disable wifi
-Open Magnet
-Enable WiFi
-Refresh Magnet

arcturus:
Closes #284

### DIFF
--- a/ios/Magnet/AppDelegate.m
+++ b/ios/Magnet/AppDelegate.m
@@ -21,7 +21,10 @@
 {
   NSURL *jsCodeLocation;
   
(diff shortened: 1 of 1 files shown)


wilsonpage:
r+

isabelrios:
Not able to reproduce the #284 bug so far. Thanks! @arcturus

PR #287, opened 2016-11-01
Add subscriptions/channels UI

wilsonpage:


android/app/src/main/java/org/mozilla/magnet/api/Utils.java
```diff
@@ -0,0 +1,158 @@
+package org.mozilla.magnet.api;
```
wilsonpage, line 1: Stolen code :)

android/app/src/main/java/org/mozilla/magnet/api/Api.java
```diff
@@ -0,0 +1,148 @@
...
+    public Api(Context context) {
+        mContext = context;
+    }
+
+    protected void mount(String namespace, Api api) {
+        routes.put(namespace, api);
+    }
+
+    private Api find(String path) {
+        for (Map.Entry<String, Api> entry : routes.entrySet()) {
```
arcturus: Why do we have this loop? Can we just check if the key exists and access to it directly right?

android/app/src/main/java/org/mozilla/magnet/api/Api.java
```diff
@@ -0,0 +1,148 @@
...
+    protected Store getCache() {
+        return Store.get(mContext);
+    }
+
+    public void post(String path, HashMap<String,Object> data, Callback callback) {
+        Api match = find(path);
+        if (match != null) match.post(path, data, callback);
+    }
+
+//    public void put(String path) {
```
arcturus: Dead code, remove it please

android/app/src/main/java/org/mozilla/magnet/api/Api.java
```diff
@@ -0,0 +1,148 @@
...
+        match.get(path, callback);
+    }
+
+    protected Store getCache() {
+        return Store.get(mContext);
+    }
+
+    public void post(String path, HashMap<String,Object> data, Callback callback) {
+        Api match = find(path);
+        if (match != null) match.post(path, data, callback);
```
arcturus: if no match we should call error

android/app/src/main/java/org/mozilla/magnet/api/Api.java
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+//    public void put(String path) {
+//        Api match = find(path);
+//        if (match != null) match.get(path);
+//    }
+
+    public void delete(String path, HashMap<String,Object> data, Callback callback) {
+        Api match = find(path);
+        if (match != null) match.delete(path, data, callback);
```
arcturus: same here, right nowif there is no route we don't know what happens

android/app/src/main/java/org/mozilla/magnet/api/Api.java
```diff
@@ -0,0 +1,148 @@
...
+        getQueue().add(request);
+    }
+
+    protected boolean hasConnectivity() {
+        ConnectivityManager cm = (ConnectivityManager) mContext.getSystemService(Context.CONNECTIVITY_SERVICE);
+        NetworkInfo activeNetwork = cm.getActiveNetworkInfo();
+        return activeNetwork != null && activeNetwork.isConnectedOrConnecting();
+    }
+
+    public interface Callback {
```
arcturus, line 194: can we change this interface to define two methods, one for success and one for error, kind or promise reject and resolve.

arcturus (commented):
Generally looking pretty good, there is one part that i'm not convinced: about removing all the cod ewe got around the subscriptions.
I'm ok on removing the native api, but we could use the current sqlite work and wrap it with the new api thing.

arcturus:
shiiiply!

r+

PR #291, opened 2016-11-02
Ios content api

arcturus:
@wilsonpage r?


ios/ApiChannels.swift
```diff
@@ -0,0 +1,39 @@
...
+    
+    let url = NSURL(string: ApiChannels.URL_STRING)
+    
+    let request = NSURLRequest(URL: url!, cachePolicy: NSURLRequestCachePolicy.ReloadIgnoringLocalCacheData, timeoutInterval: 30)
+    
+    Alamofire.request(request).responseJSON { response in
+      guard (response.result.value != nil) else {
+        callback.onError("Error performing request \(ApiChannels.URL_STRING)")
+        return
+      }
```
wilsonpage, line 31: NIT: Linebreak after block

ios/ApiMagnetReact.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+
+import Foundation
+
+@objc(ApiMagnetReact) class ApiMagnetReact: NSObject {
+  private let api: ApiMagnet = ApiMagnet()
+  
+  @objc func get(path: String,
+                 resolve: RCTPromiseResolveBlock,
+                 reject: RCTPromiseRejectBlock) {
```
wilsonpage: NIT: Don't think we use this argument formating style anywhere else in the swift code?

ios/Api.swift
```diff
@@ -0,0 +1,84 @@
...
+import Foundation
+import SwiftyJSON
+
+protocol Api {
+  func get(path: String, callback: ApiCallback)
+  func post(path: String, data: NSDictionary, callback: ApiCallback)
+  func put(path: String, data: NSDictionary, callback: ApiCallback)
+  func delete(path: String, data: NSDictionary, callback: ApiCallback)
+  func mount(path: String, api: Api)
+}
```
wilsonpage, line 18: Do we need this `protocol`?

ios/ApiMagnetReact.swift
```diff
@@ -0,0 +1,64 @@
...
+    }))
+  }
+  
+  @objc func post(path: String,
+                  data: NSDictionary,
+                  resolve: RCTPromiseResolveBlock,
+                  reject: RCTPromiseRejectBlock) {
+    
+    api.post(path, data: data, callback: ApiCallback(success: { (result) in
+      resolve(result.rawValue)
```
wilsonpage: Indentation

ios/ApiMagnetReact.swift
```diff
@@ -0,0 +1,64 @@
...
+        reject("get_error", "Error resolving \(path)", err)
+    }))
+  }
+  
+  @objc func post(path: String,
+                  data: NSDictionary,
+                  resolve: RCTPromiseResolveBlock,
+                  reject: RCTPromiseRejectBlock) {
+    
+    api.post(path, data: data, callback: ApiCallback(success: { (result) in
```
wilsonpage: Previously didn't use braces on argument. Can we be consistent?

ios/ApiMagnetReact.swift
```diff
@@ -0,0 +1,64 @@
...
+    }))
+  }
+  
+  @objc func put(path: String,
+                 data: NSDictionary,
+                 resolve: RCTPromiseResolveBlock,
+                 reject: RCTPromiseRejectBlock) {
+    
+    api.put(path, data: data, callback: ApiCallback(success: { (result) in
+      resolve(result.rawValue)
```
wilsonpage: Indentation

ios/ApiMagnetReact.swift
```diff
@@ -0,0 +1,64 @@
...
+    }))
+  }
+  
+  @objc func delete(path: String,
+                    data: NSDictionary,
+                    resolve: RCTPromiseResolveBlock,
+                    reject: RCTPromiseRejectBlock) {
+    
+    api.delete(path, data: data, callback: ApiCallback(success: { (result) in
+      resolve(result.rawValue)
```
wilsonpage: Indentation

ios/Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
 github "Alamofire/Alamofire" "3.5.1"
 github "stephencelis/SQLite.swift" "8b95a0c4a884f35e5102cd306f634680255473bb"
 github "SwiftyJSON/SwiftyJSON" "54017d514a87b2b855b407131292c510cdeb65f8"
-github "mozilla-magnet/magnet-scanner-ios" "0.3.2"
+github "mozilla-magnet/magnet-scanner-ios" "0.3.3"
```
wilsonpage, line 4: What changes does this pull in?

ios/Magnet/AppDelegate.m
```diff
@@ -22,9 +22,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   NSURL *jsCodeLocation;
   
   // Load from the server
-  //jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
+  jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
   // Load from local bundle
-  jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
+  //jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
```
wilsonpage: Did you mean to change this?

ios/RequestStoreSQLite.swift
```diff
@@ -0,0 +1,77 @@
...
+  }
+  
+  func delete(_url: String) {
+    let item = requestStore.filter(url == _url)
+    try! db.run(item.delete())
+  }
+  
+  func get(_url: String) -> JSON? {
+    let query = requestStore.filter(url == _url).limit(1)
+
```
wilsonpage: NIT: Unnecessary line break

ios/ApiSubscriptions.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Created by Francisco Jordano on 02/11/2016.
+//  Copyright © 2016 Mozilla. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+class ApiSubscriptions: ApiBase {
+  
+  private static let PATH = "content://subscriptions"
```
wilsonpage, line 14: Unused var?

ios/System.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Copyright © 2016 Mozilla. All rights reserved.
+//
+
+import Foundation
+
+
+import SystemConfiguration
+
+class System {
+  class func connectedToNetwork() -> Bool {
```
wilsonpage: Are you sure you can't use the `static` keyword here? A lot more readable.

ios/ApiChannels.swift
arcturus: Sure

ios/ApiMagnetReact.swift
arcturus: Sure, with the longer names in the previous signature was a pain, but now we can go back to normal formatting

ios/Api.swift
arcturus: Unfortunately in swift there is no abstract classes, but the protocol is the interface. Then we use the ApiBase as a superclass for the rest of the apis.

Strictly not necessary but like to keep it.

ios/Magnet/AppDelegate.m
arcturus: Definitely not on purpose, good catch!

ios/System.swift
arcturus: we can use static there, that code is actually taken from stackoverflow and i don't think we will need to overwrite it. The difference between static and class in static methods is that if you have static you won't be able to over write in a subclass.

ios/ApiSubscriptions.swift
arcturus: yup, from previous implementation

PR #301, opened 2016-11-03
Subscribed notifications (android)

wilsonpage:


wilsonpage:
Depends on https://github.com/mozilla-magnet/magnet-scanner-android/pull/19

@arcturus r?

android/app/src/main/java/org/mozilla/magnet/notifications/ReceiverNotificationDelete.java
```diff
@@ -0,0 +1,48 @@
...
+import android.util.Log;
+
+public class ReceiverNotificationDelete extends BroadcastReceiver {
+    private final String TAG = "NotificationDelete";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.d(TAG, "on receive");
+        String action = intent.getAction();
+//        if (!action.equals("notification-dismiss")) { return; }
```
arcturus: nit: dead code

android/app/src/main/java/org/mozilla/magnet/notifications/ReceiverNotificationDelete.java
```diff
@@ -0,0 +1,48 @@
...
+import android.support.v4.content.LocalBroadcastManager;
+import android.util.Log;
+
+public class ReceiverNotificationDelete extends BroadcastReceiver {
+    private final String TAG = "NotificationDelete";
+
+    @Override
+    public void onReceive(Context context, Intent intent) {
+        Log.d(TAG, "on receive");
+        String action = intent.getAction();
```
arcturus, line 18: Do we need this variable?

android/app/src/main/java/org/mozilla/magnet/MyAppPackage.java
```diff
@@ -22,6 +23,7 @@
         modules.add(new MagnetScannerReact(reactContext));
         modules.add(new PromptBluetoothReact(reactContext));
         modules.add(new ApiMagnetReact(reactContext));
+        modules.add(new NotificationListenerReact(reactContext));
```
arcturus, line 26: Why do we need to expose this to JS land?

android/app/src/main/java/org/mozilla/magnet/api/Api.java
```diff
@@ -101,14 +107,23 @@ private RequestQueue getQueue() {
...
     /**
      * Request a particular URL and parse the result as `JSONArray`.
      *
      * @param url
      * @param callback
      */
-    protected void requestJsonArray(String url, final Callback callback) {
-        request(Request.Method.GET, url, new Callback() {
+    protected void requestJsonArray(int method, String url, String body, final Callback callback) {
+        request(method, url, body, new Callback() {
```
arcturus, line 126: It's not clear to me, cause a body should be there just for the post request right? this could be a bit ambiguous IMO.

android/app/src/main/java/org/mozilla/magnet/notifications/ImageNotification.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    static ImageNotification create(Context context, MagnetScannerItem item, int id) {
+        ImageNotification notification = new ImageNotification(context, item, id);
+        notification.execute();
+        return notification;
+    }
+
+    @Override
+    protected Map<String,Bitmap> doInBackground(Void... voids) {
```
arcturus, line 35: Do we need to overwrite this method? Is not the same than the super class? Or maybe i'm not seeing a tiny difference.

android/app/src/main/java/org/mozilla/magnet/notifications/MagnetNotification.java
```diff
@@ -0,0 +1,118 @@
...
+
+        return PendingIntent.getActivity(
+                mContext,
+                uniqueRequestCode,
+                launchIntent,
+                PendingIntent.FLAG_UPDATE_CURRENT);
+    }
+
+    PendingIntent createDismissIntent() {
+        return ReceiverNotificationDelete.createIntent(mContext, mId);
```
arcturus, line 90: Should we notify stats for a notification dismissed?

arcturus (commented):
First pass, looking pretty good, can you address the questions?

android/app/src/main/java/org/mozilla/magnet/notifications/ReceiverNotificationDelete.java
wilsonpage: Will remove

android/app/src/main/java/org/mozilla/magnet/MyAppPackage.java
wilsonpage: So that we can track notification-dismiss in JS analytics. Once we have native analytics it would be best to be done there as if the app is closed we can't track.

android/app/src/main/java/org/mozilla/magnet/notifications/ImageNotification.java
wilsonpage: You're right, I should remove this logic from the base class.

android/app/src/main/java/org/mozilla/magnet/notifications/MagnetNotification.java
wilsonpage: We don't have any way of collecting metrics from native code yet, but JS land should be notified if it's running and this is the same code path we collect the metrics from in master today.

PR #311, opened 2016-11-07
Ensure that the system level alert permission is removed from release builds

### LINKED ISSUE #309: App still asks for 'draw over other apps' permission

samgiles:
Closes #309

@wilsonpage r?

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -121,12 +121,7 @@ android {
         }
     }
     buildTypes {
-        debug {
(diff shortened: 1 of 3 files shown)


wilsonpage:
@samgiles sweet!!
- So does the release manifest get merged with the default manifest?
- Are you able to strip out the old gradle flag that wasn't actually working.

wilsonpage:
r+ after comments shhhhhhip.ly

Issue #296, opened 2016-11-03
Add ability to call Google Analytics from native code


wilsonpage:
As notifications become more advanced more of our code has to run on the native side. We need to be able to call Google Analytics from native code so that we can learn from user interactions with notifications when the app is closed.

wilsonpage:
We can either run this a parallel with `react-native-google-analytics-bridge` or replace it with our own wrapper that can be called from JS or Native code.

arcturus:
Do you thikn it makes sense to send a PR to `react-native-google-analytics-bridge` to allow native use of the api?

wilsonpage:
That would be cool.

samgiles:
Gonna break this down into three tasks
- [x] Expose user preferences to native code through the `Api` mechanism.
- [ ] Expose the native GA libraries from the `react-native-google-analytics-bridge`
- [ ] Add calls to analytics from notifications and background tasks.

wilsonpage:
@samgiles ... like a boss 😎

samgiles:
Closed in #349

PR #319, opened 2016-11-08
Part 1 of #296: Expose preferences to the native side

samgiles:
Part 1 for #296.

Removes the dependency on AsyncStorage \o/.

I've also tried to make the backend API for Preferences more generic than booleans only.

r? @wilsonpage Android: Also see how there's a mutex lock over the delete and update methods, should prevent weird races.  We might want to apply that technique in the subscriptions API too. 

r? @arcturus iOS  I'm not sure about potential concurrency issues here, do we need a mutex marshalling access through the API for remove and update?


ios/ApiPreferences.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Copyright © 2016 Mozilla. All rights reserved.
+//
+
+import Foundation
+import SwiftyJSON
+
+class ApiPreferences: ApiBase {
+  
+  private static let PATH = "content://preferences"
+  private var preferences: Preferences!
```
arcturus: Oh! I though we were going to use the RequestStore, where we can store json directly, but tbh, I like this idea of having sqlite stores :)

ios/Magnet/AppDelegate.m
```diff
@@ -20,17 +20,17 @@ @implementation AppDelegate
...
   NSURL *jsCodeLocation;
-  
+
   // Load from the server
-  //jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
+  jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
   // Load from local bundle
-  jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
-  
+  //jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
```
arcturus: Don't commit this change ;)

ios/Preferences.swift
```diff
@@ -0,0 +1,30 @@
...
+class Preferences: NSObject {
+  private var db: PreferencesDB!
+  
+  override init() {
+    super.init()
+    db = PreferencesDB.get()
+  }
+  
+  func remove(prefKey: String) -> Bool {
+    return db.remove(prefKey)
```
arcturus: Regarding your question about the locking, we could start doing transactions, they run in their own queue.

Is something that I didn't do for the other sqlite implementations, but I'm planning to add:

https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Index.md#transactions-and-savepoints

More info about thread safe operations:
https://github.com/stephencelis/SQLite.swift/blob/master/Documentation/Index.md#thread-safety

ios/PreferencesDB.swift
```diff
@@ -0,0 +1,88 @@
...
+    return singleton
+  }
+  
+  private init() {
+    let path = NSSearchPathForDirectoriesInDomains(
+      .DocumentDirectory,
+      .UserDomainMask,
+      true).first!
+    
+    db = try! Connection("\(path)/history.sqlite3")
```
arcturus: So far we have been using different sqlite dbs for each data type, this will create a table into the history db.

samgiles: I noticed this is what the subscriptions was doing and thought you were heading for one DB?

ios/ApiPreferences.swift
samgiles: Ohhh, the RequestStore would've been simpler :P, but OK.

ios/Preferences.swift
samgiles: Ah, ok, this would help. I used a transaction for the update or insert logic.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiMagnet.java
```diff
@@ -16,5 +16,6 @@ public ApiMagnet(Context context) {
         mount("channels", new ApiChannels(context));
         mount("subscriptions", new ApiSubscriptions(context));
         mount("metadata", new ApiMetadata(context));
+        mount("preferences", new ApiPreferences(context));
```
wilsonpage, line 19: @samgiles off topic a little: Interested how you imagine these user-state routes eventually persisting to magnet-service, are they in an alright state to do so?

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
```diff
@@ -0,0 +1,106 @@
...
+    public void get(String aPath, Callback aCallback) {
+        aCallback.resolve(getCache().getJsonObject(PREFERENCES_STORE_KEY));
+    }
+
+    /**
+     * Update or set a preference given a Map with structure:
+     * `{ "pref_key": "string", "value": "string" }`
+     */
+    @Override
+    public void post(String aPath, Object aData, Callback aCallback) {
```
wilsonpage, line 39: What's a `a` prefix convention?

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
```diff
@@ -0,0 +1,106 @@
...
+        if (!(aData instanceof Map)) {
+            aCallback.reject("Data must be a Map type");
+            return;
+        }
+
+        // Lock everything between getting the JSON object from the cache and
+        // writing it back - a synchronize on the method isn't enough as the
+        // resource is used in the delete method too.
+        mWriteLock.lock();
```
wilsonpage, line 50: What is the scope of this lock? Can you define what a lock does?

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
```diff
@@ -0,0 +1,106 @@
...
+
+        Map data = (Map) aData;
```
wilsonpage: Why is this reassigned?

ios/ApiPreferences.swift
wilsonpage: Agree, this patch could be a lot simpler if we re-use our simple key/value `RequestStore`.

ios/PreferencesDB.swift
wilsonpage: For simple JSON data we should just use the `RequestStore`, although the term 'request' is a little misleading, it's really just our own key/value store, like `AsyncStorage`.

wilsonpage (commented):
Looks good, nice work!
- Android looks perfect, I just need to learn more about the locking stuff you did. New to me! :)
- iOS patch is a lot bigger because it doesn't use the generic key/value cache. I'd rather we keep all data stores as dumb as possible until the need for a dedicated DB/Table arises.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
samgiles: 'aData' is type Object, we need a 'Map', the check to ensure it's an instance of 'Map' happens before we acquire the lock.  Could move the cast up outside of the locked area too.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
samgiles: "a" for arguments, "m" for "member", it's the convention Fennec uses.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
samgiles: Unfortunately in Java you can't scope a lock, there isn't, as far as I know, a way to do RAII (Resource acquisition is initialization) to scope the lock and have it release automatically. Although it looks like there might be a way to do this in Java 7, I'm not sure if it works with Locks, ~~I should try~~–it doesn't.

So given this, the code between calling 'lock' and 'unlock' is the scope, aka the critical section.

A lock is a mutual exclusion mechanism, it ensures that only one thread is able to execute the code between acquiring the lock, and releasing the lock.  The code block protected by a lock is called the critical section.  If a thread "holds the lock", no other thread can enter the critical section until it acquires the lock-when the other thread releases the lock.  It provides a way of linearising access to a resource and ensuring certain operations on shared memory are atomic.  An example directly applicable here is, we don't want to add an entry, and then read it back to find that during the add operation, another thread actually deleted it.

Java has the 'synchronised' keyword that will mutex lock a single method.  But we need to lock across a couple of methods here.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiMagnet.java
samgiles: Yes seems to make sense so far.

samgiles:
@wilsonpage How's this?

ios/ApiPreferences.swift
```diff
@@ -10,34 +10,62 @@ import Foundation
...
-    
-    var dict = Dictionary<String, String>()
-    
-    dict.updateValue(data["value"]!.description, forKey: data["pref_key"] as! String)
-    
-    preferences.update(dict)
-    let json = JSON(dict)
+    var json = JSON("{}");
+
+    dispatch_sync(lockQueue) {
```
wilsonpage, line 35: Some brief comments about how/why of this lock techniques would be nice :)

ios/ApiPreferences.swift
```diff
@@ -10,34 +10,62 @@ import Foundation
...
+        json = storeJson
+        store.setJSON(ApiPreferences.PATH, value: storeJson)
+      }
+    }
+
     callback.onSuccess(json)
   }
 }
+
+func tryAsBooleanString(value: String) -> String {
```
wilsonpage, line 75: Comments to explain what this is for.

ios/ApiPreferences.swift
```diff
@@ -10,34 +10,62 @@ import Foundation
...
-    let json = JSON(dict)
+    var json = JSON("{}");
+
+    dispatch_sync(lockQueue) {
+
+      if let storedJson = store.getJSON(ApiPreferences.PATH) {
+        json = storedJson
+      }
+
+      json[data["pref_key"] as! String] = JSON(tryAsBooleanString(data["value"]!.description))
```
wilsonpage: What is `.description`?

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
wilsonpage: Thanks for the explanation 😃

wilsonpage:
@samgiles just few comments needed in places. But then ship.ly 👍

PR #320, opened 2016-11-08
Downgrade google-play-services to 9.6.1

wilsonpage:
- 9.8.0 is too recent meaning we need to prompt many devices to upgrade (friction)
- Also 9.8.0 doesn't work on the android emulators yet
- Found out that react-native-google-analytics-bridge wasn't specifying an explicit version which was causing a strange merging of versions. We're using gradle's exclude feature to override this transitive dependency.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -146,12 +146,22 @@ android {
 }
 
 dependencies {
-    compile project(':react-native-google-analytics-bridge')
     compile project(':react-native-linear-gradient')
     compile project(':magnet-scanner-android')
-    compile fileTree(dir: "libs", include: ["*.jar"])
-    compile "com.android.support:appcompat-v7:23.0.1"
-    compile "com.facebook.react:react-native:+"  // From node_modules
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    compile 'com.android.support:appcompat-v7:23.0.1'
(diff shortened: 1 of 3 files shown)


wilsonpage:
@arcturus r?

android/app/build.gradle
```diff
@@ -146,12 +146,22 @@ android {
...
-    compile "com.facebook.react:react-native:+"  // From node_modules
+    compile fileTree(dir: 'libs', include: ['*.jar'])
+    compile 'com.android.support:appcompat-v7:23.0.1'
+    compile 'com.facebook.react:react-native:+'  // From node_modules
+    compile 'com.bugfender.sdk:android:0.+'
+
+    // don't include google-play-services 'transitive'
+    // dependency so that we can be sure that we get
+    // the exact version included in magnet-scanner-android
+    compile(project(':react-native-google-analytics-bridge')) {
```
wilsonpage, line 159: This was what was killing us.

wilsonpage:
Depends on https://github.com/mozilla-magnet/magnet-scanner-android/pull/22

arcturus:
Nice catch, shiiip it!

arcturus:
Ugh travis complain :S

wilsonpage:
Fixing

PR #329, opened 2016-11-10
Move permissions logic into native code (closes #323)

### LINKED ISSUE #323: Move prompts out of JS code
It's becoming clear that specific hardware/feature availability prompts are very platform specific. For example in Android apps must perform checks and prompts manually, whereas IIUC iOS does this automatically.

JS code should be reserved for logic that must be run for both platforms.

wilsonpage:


wilsonpage:
@arcturus f/r?

lib/scanner/index.js
```diff
@@ -41,43 +39,26 @@ class Scanner extends Emitter {
...
-        // it's an error if the user denied the bluetooth prompt
-        if (!enabled) return Promise.reject('bt-disabled');
-
-        // actually start scanning now
-        MagnetScanner.start();
-        this.started = true;
-
+    return MagnetScanner.start()
+      .then(() => {
+        debug('started');
```
wilsonpage, line 51: Here we've effectively made the client dumber and the native side smarter. This is because the native side knows what its platform's dependencies are before the scanner can run, whereas the client is too abstract to make such decisions.

I found that we had non-trivial protection logic in both JS and native, it's best if we only have to maintain this stuff in one place. If the client attempts to call `.start()` when native side is 'busy', the promise will reject. But the client is aware that some errors are less serious than others.

lib/scanner/index.js
```diff
@@ -41,43 +39,26 @@ class Scanner extends Emitter {
...
   }
 
   stop() {
     debug('stop');
-    delete this.starting;
-    if (!this.started) return;
     MagnetScanner.stop();
     this.stopExpireCheck();
-    this.listener.remove();
-    this.started = false;
```
wilsonpage, line 80: As above, removing state from this component and moving it to the native side simplifies the client.

lib/scanner/index.js
```diff
@@ -116,28 +97,6 @@ class Scanner extends Emitter {
...
-        }
-
-        return this.showWarning()
-          .then(retry => {
-            this.checking = null;
-            if (!retry) return false;
-            return this.checkBluetooth();
-          });
-      });
-  }
```
wilsonpage, line 139: Checking bluetooth is an Android only issue (IIUC), so we've moved this into Android only code, making usre the user is prompted before the scanner runs.

ios/MagnetScannerReact.m
```diff
@@ -0,0 +1,13 @@
...
+
+RCT_EXTERN_METHOD(
+                  start:(RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject)
+
+RCT_EXTERN_METHOD(
+                  stop:(RCTPromiseResolveBlock)resolve
+                  reject:(RCTPromiseRejectBlock)reject)
+
+@end
```
wilsonpage, line 13: Changed name of file to match convention of our other modules.

wilsonpage: Both `start()` and `stop()` return promises now as we have given native code responsibility to manage device dependencies before scanner can run. For example prompting the user for bluetooth is a slow async process which blocks the magnet-scanner-android from running. The promise doesn't resolve until these permission dependencies have been checked and prompted.

wilsonpage:
@samgiles you free to r?

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
```diff
@@ -37,28 +53,118 @@ public String getName() {
...
 
     /**
-     * Start the ScannerService scanning.
+     * Start the scanner.
+     *
+     * Before the scanner runs we check and prompt the user
+     * for all the permissions required to run the scanner.
+     *
+     * We don't yet block the scanner from running if
+     * the user doesn't accept accept all the prompts,
```
samgiles: NIT: typo "accept accept"

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
```diff
@@ -37,28 +53,118 @@ public String getName() {
...
+            Log.e(TAG, "can't check when app in background");
+            callback.onPermissionChecksError(new Error("activity null"));
+            return;
+        }
+
+        mPermissionChecker.check(activity, callback);
+    }
+
+    private boolean needsPermissionsCheck() {
+        long MIN_TIME = TimeUnit.SECONDS.toMillis(10);
```
samgiles: NIT: Can we get rid of the magic number `10` into a constant?

You could make this whole thing a static member 
`private static final long MIN_TIME = TimeUnit.SECONDS.toMillis(10)`

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,44 @@
...
+
+    protected void done() {
+        mListener.onCheckDone(this);
+    }
+
+    protected Activity getActivity() {
+        return mActivity;
+    }
+
+    void awaitingResponse() {
```
samgiles: NIT: Explicit `private|protected|public`

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,44 @@
...
+abstract class Check {
+    private Activity mActivity;
+    private PermissionChecker mListener;
+
+    Check(PermissionChecker listener, Activity activity) {
+        mActivity = activity;
+        mListener = listener;
+    }
+
+    abstract int getId();
```
samgiles: NIT: explicit `private|protected|public`.

And the line below :)

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,44 @@
...
+import android.app.AlertDialog;
+import android.app.Dialog;
+import android.content.DialogInterface;
+import android.content.pm.PackageManager;
+import android.support.v4.content.ContextCompat;
+
+/**
+ * Created by wilsonpage on 09/11/2016.
+ */
+abstract class Check {
```
samgiles: NIT: Explicit `private|public`

Making scope qualifiers explicit just makes it easier to understand the external interface. Particularly if you're not used to Java's defaults (like me 👍 )

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,44 @@
...
+
+    protected Activity getActivity() {
+        return mActivity;
+    }
+
+    void awaitingResponse() {
+        mListener.onCheckAwaitingResponse(this);
+    }
+
+    boolean hasPermission(String permission) {
```
samgiles: NIT: Explicit `private|protected|public`

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
samgiles: There are a few places in this file and its implementers.

android/app/src/main/java/org/mozilla/magnet/permissions/CheckBluetooth.java
```diff
@@ -0,0 +1,59 @@
...
+        awaitingResponse();
+    }
+
+    public void onResponse(int result) {
+        boolean success = result == Activity.RESULT_OK;
+        done();
+    }
+
+    private boolean hasBluetooth() {
+        return BluetoothAdapter.getDefaultAdapter() != null;
```
samgiles: Seems like we've got two methods of getting a `BluetoothAdapter`. Here we use a static method, and in `bluetoothEnabled` we get the `BluetoothManager` and get the adapter from there.  We should consolidate these functions to avoid duplication/inadvertently checking the wrong thing.

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,44 @@
...
+import android.content.DialogInterface;
+import android.content.pm.PackageManager;
+import android.support.v4.content.ContextCompat;
+
+/**
+ * Created by wilsonpage on 09/11/2016.
+ */
+abstract class Check {
+    private Activity mActivity;
+    private PermissionChecker mListener;
```
samgiles: If this is only a listener, make the type the interface `CheckListener`.  It's generally better to use the interface rather than the implementer because it decouples you from the `PermissionChecker` and you only expose the type that is required.

samgiles (commented):
Looks good!  I like how you managed the concurrent control flow in the end, nice and simple.

 A few nits/comments, r+  once they're addressed.

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
arcturus: +1

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
```diff
@@ -40,25 +56,98 @@ public String getName() {
...
+    }
+
+    @Override
+    public void onActivityResult(Activity activity, int requestCode, int resultCode, Intent data) {
+        PermissionChecker.onResponse(requestCode, resultCode);
+    }
+
+    @Override
+    public void onNewIntent(Intent intent) {
+        // noop
```
arcturus, line 156: Why do we need to override this method?

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
```diff
@@ -40,25 +56,98 @@ public String getName() {
...
+        if (activity == null) {
+            Log.e(TAG, "can't check when app in background");
+            callback.onPermissionChecksError(new Error("activity null"));
+            return;
+        }
+
+        mPermissionChecker.check(activity, callback);
+    }
+
+    private boolean needsPermissionsCheck() {
```
arcturus, line 144: I though the properties starting and started were going to be used to control that we don't need to check permissions again, but seems we are doing it based on time, don't you think this can cause races?

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,55 @@
...
+import android.app.AlertDialog;
+import android.app.Dialog;
+import android.content.DialogInterface;
+import android.content.pm.PackageManager;
+import android.support.v4.content.ContextCompat;
+
+/**
+ * Created by wilsonpage on 09/11/2016.
+ */
+abstract class Check {
```
arcturus: +1 to all the comments from @samgiles about specifying the visibility of the class and methods, otherwise the visibility is just for the package.

Also I would rename the class, `Check` is too generic. I know we have it under the package permissions, but I like verbosity in class names.

android/app/src/main/java/org/mozilla/magnet/permissions/CheckBluetooth.java
```diff
@@ -0,0 +1,59 @@
...
+import android.app.Activity;
+import android.bluetooth.BluetoothAdapter;
+import android.bluetooth.BluetoothManager;
+import android.content.Context;
+import android.content.Intent;
+
+/**
+ * Created by wilsonpage on 09/11/2016.
+ */
+class CheckBluetooth extends Check {
```
arcturus: nit: visibility

android/app/src/main/java/org/mozilla/magnet/permissions/CheckBluetooth.java
```diff
@@ -0,0 +1,59 @@
...
+    private static int ID = 10;
```
arcturus: nit: make it final, it's a constant right?

android/app/src/main/java/org/mozilla/magnet/permissions/CheckBluetooth.java
```diff
@@ -0,0 +1,59 @@
...
+    CheckBluetooth(PermissionChecker listener, Activity activity) {
+        super(listener, activity);
+    }
+
+    @Override
+    public int getId() {
+        return ID;
+    }
+
+    void check() {
```
arcturus: nit: visibility

android/app/src/main/java/org/mozilla/magnet/permissions/CheckBluetooth.java
```diff
@@ -0,0 +1,59 @@
...
+        if (!hasBluetooth()) {
+            done();
+            return;
+        }
+
+        if (bluetoothEnabled()) {
```
arcturus: nit: this two `if` can be consolidated into a single one

android/app/src/main/java/org/mozilla/magnet/permissions/CheckGooglePlayServices.java
```diff
@@ -0,0 +1,54 @@
...
+
+import android.app.Activity;
+
+import com.google.android.gms.common.ConnectionResult;
+import com.google.android.gms.common.GoogleApiAvailability;
+
+/**
+ * Created by wilsonpage on 09/11/2016.
+ */
+class CheckGooglePlayServices extends Check {
```
arcturus: nit: visibility

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
```diff
@@ -0,0 +1,55 @@
...
+    private PermissionChecker mListener;
+
+    Check(PermissionChecker listener, Activity activity) {
+        mActivity = activity;
+        mListener = listener;
+    }
+
+    abstract int getId();
+    abstract void onResponse(int response);
+    abstract void check();
```
arcturus: Seems the `check` method needs to be run in the ui thread right? Since it needs usually an activity. Well, either the `check` method or all the checkers.

Maybe we can add an assertion when we create a checker object to be sure we run them in the UI thread, we can check if we are in the ui thread with:
`Looper.myLooper() == Looper.getMainLooper()`

android/app/src/main/java/org/mozilla/magnet/permissions/CheckLocationPermission.java
```diff
@@ -0,0 +1,50 @@
...
+
+import android.Manifest;
+import android.app.Activity;
+import android.content.pm.PackageManager;
+import android.support.v4.app.ActivityCompat;
+
+/**
+ * Created by wilsonpage on 09/11/2016.
+ */
+class CheckLocationPermission extends Check {
```
arcturus: nit: guess what ;P

android/app/src/main/java/org/mozilla/magnet/permissions/PermissionChecker.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+public class PermissionChecker implements CheckListener {
+    private static Map<Integer,Check> checksAwaitingResponse = new HashMap<>();
+    private Stack<Check> checks = new Stack<>();
+    private PermissionCheckerCallback mCallback;
+    private boolean mChecking = false;
+
+    public void check(Activity activity, PermissionCheckerCallback callback) {
+        if (mChecking) return;
```
arcturus, line 20: my eyes, my eyes :P

arcturus:
Looking pretty good to me, please take a look to the suggestions and shiiiip it!

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
wilsonpage: Android studio keeps prompting me to remove redundant scope qualifiers. But happy to be more explicit.

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
wilsonpage: It's a requirement of `ActivityEventListener` interface, but we don't actually make use of it.

android/app/src/main/java/org/mozilla/magnet/MagnetScannerReact.java
wilsonpage: This is a really tricky one. We start/stop the scanner via onResume/onPause. When we show dialogs during the permissions checking it triggers `onPause` when they are show and `onResume` when they are dismissed. This causes the scanner `.start()` and `.stop()` to be called repeatedly during permissions checking.

You're right that the `mStarting` flag will be `true` during this time, but when the final dialog is dismissed it can cause the permissions check cycle to start again. This is not an ideal solution, but I'm out of ideas.

android/app/src/main/java/org/mozilla/magnet/permissions/Check.java
wilsonpage: FYI Android studio told me to remove unnecessary scope declarations when something is `Package local`.

wilsonpage: screen shot 2016-11-16 at 08 19 44

android/app/src/main/java/org/mozilla/magnet/permissions/CheckGooglePlayServices.java
wilsonpage: I don't think I want to make this `public` as it means the class is reachable outside of the `permissions` package which isn't the intention.

wilsonpage:
Landed. Happy to make any further suggestions as followups.

PR #334, opened 2016-11-14
Implementing external logging system for debug builds in ios

arcturus:
@wilsonpage without having to add the pod, which was messing with the project layout.


arcturus:
r=me

PR #338, opened 2016-11-15
Deep linking (closes #331)

### LINKED ISSUE #331: Deep linking to POI View (android)
When a notification is tapped the app should open directly into the corresponding POI view.

wilsonpage:


wilsonpage:
@arcturus r? (ping-pong sorry)

lib/store/constants.js
```diff
@@ -0,0 +1,6 @@
+
+module.exports = {
+  INCOMPLETE: 'incomplete',
```
wilsonpage, line 3: `FETCHING`

lib/store/reducer.js
```diff
@@ -4,18 +4,24 @@
...
+  ERROR,
+} = require('./constants');
+
 const initialState = {
-  items: [],
-  openItem: {
-    isFetching: false,
-    value: null,
-  },
+  items: {},
```
wilsonpage, line 18: `itemsCache`?

lib/views/item/item.js
```diff
@@ -2,8 +2,8 @@
  * Dependencies
```
wilsonpage: Change filename to `item-detail.js`

lib/views/item/container.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
```
wilsonpage: -> `item-container.js`

lib/views/list/item.js
```diff
@@ -5,8 +5,8 @@
  */
```
wilsonpage: `list-item.js`

lib/views/item/scene.js
```diff
@@ -4,28 +4,42 @@
...
-  navigator: React.PropTypes.object.isRequired,
-  itemUrl: React.PropTypes.string,
-  item: React.PropTypes.object,
-  setOpenItem: React.PropTypes.func,
+  navigator: PropTypes.object.isRequired,
+  itemUrl: PropTypes.string,
+  openItem: PropTypes.string,
+  items: PropTypes.object,
+  setOpenItem: PropTypes.func,
+  dispatch: PropTypes.func.isRequired,
```
wilsonpage: Maybe can remove some of these?

lib/views/list/scene.js
```diff
@@ -139,30 +159,31 @@ ListScene.propTypes = {
  * @param  {ReduxStore} store
  * @return {Object}
  */
-function mapStateToProps(state) {
-  const { sortByDistance, showDistance } = state.userFlags;
-  const items = state.itemsNearby
-    .map(url => resolveItem(url))
-    .filter(item => !!item);
+function mapStateToProps({ items, itemsNearby, userFlags, indicateScanning }) {
```
wilsonpage: Remove `items`

wilsonpage: Ignore

ios/ApiMetadata.swift
```diff
@@ -35,20 +35,25 @@ class ApiMetadata: ApiBase {
   //
   //  data["objects"] = elems
```
wilsonpage, line 36: Remove comment

wilsonpage: *Update

arcturus:
Follow ups:
- Make a service that returns both metadata + beacon data. Outside of metadata server, remove the api in js land that do the mixing (https://github.com/mozilla-magnet/magnet-services/issues/30)
- Batch requests about objects (https://github.com/mozilla-magnet/magnet-client/issues/347).
- onLost logic for ios scanners. (https://github.com/mozilla-magnet/magnet-scanner-ios/issues/13)
- Cache of items, at some point clear old items, but when the items get complicated. Now it's not a problem (https://github.com/mozilla-magnet/magnet-client/issues/346)
- Rename files to match class name.
- Calculate distance on the client from strength or lat lon, but we won't update the model that often, on the native side we should send a power mashed up (https://github.com/mozilla-magnet/magnet-scanner-ios/issues/12, https://github.com/mozilla-magnet/magnet-scanner-android/issues/25)
- When launching the app directly from deep linking dont show the list of items (https://github.com/mozilla-magnet/magnet-client/issues/348)

arcturus:
@wilsonpage after rebasing to latest master (it has a patch from @samgiles for react-native version) it does kind of work in iphone.

It at least tryies to load the view, it loads the basic, and then the app dies cause an exception in the new postArray method, need to debug more to understand whats going on.

arcturus:
More feedback about what's going on.

It's important to take into account the environment where I am, surronded by beacons that we constantly receive, so I have hundreds of calls to app render scene home in a matter of 10 seconds. Either that or we are constantly triggering model change, and also calling fetchItemIfNeeded even if it's going to end soon, we still have an enormous number of calls to that function.

That is making the animation to open the item to be very slow and to compite for resources. When we load an item that is in the cache everything seems to work ok.

arcturus:
Yes, I can confirm that in the simulator without so many update cause of btle beacons everything works way smoother.

We can consider this like an aggressive environment, but we should think on adding some throttling to that update coming from BLE.

wilsonpage:
@arcturus:
- Fixed perf issue caused by transitions being hammered by react re-render
- Tested deep links working on real ios device
- Ready to land

arcturus:
Shiiiiip it once travis is green :)

PR #339, opened 2016-11-17
Fix failing upgrade

samgiles:
Don't upgrade minor version.


samgiles:
r+ from me (am I doing it right?).

PR #341, opened 2016-11-17
Geofencing

arcturus:
@wilsonpage  this is the PR for the geofencing using regions.


ios/Cartfile
```diff
@@ -1,2 +1,2 @@
 github "stephencelis/SQLite.swift" "swift-2.3"
-github "mozilla-magnet/magnet-scanner-ios" ~> 0.3.4
+github "arcturus/magnet-scanner-ios" "betterscanning"
```
arcturus: This will change to a proper tag once https://github.com/mozilla-magnet/magnet-scanner-ios/pull/11 lands.

ios/Cartfile.resolved
```diff
@@ -1,4 +1,4 @@
 github "Alamofire/Alamofire" "3.5.1"
 github "stephencelis/SQLite.swift" "8b95a0c4a884f35e5102cd306f634680255473bb"
 github "SwiftyJSON/SwiftyJSON" "54017d514a87b2b855b407131292c510cdeb65f8"
-github "mozilla-magnet/magnet-scanner-ios" "0.3.4"
+github "arcturus/magnet-scanner-ios" "49737237c4528fda455bd9642a6ea1bdbfe8f42b"
```
arcturus: Same this, will get updated with the final commit.

ios/History.swift
```diff
@@ -38,12 +38,21 @@ class History {
...
+    #if DEBUG
+      let earlyDate = NSCalendar.currentCalendar().dateByAddingUnit(
+        .Minute,
+        value: -1,
+        toDate: NSDate(),
+        options: [])
+      return db.getSince(url, sinceDate: earlyDate!)
+    #else
+      let earlyDate = NSCalendar.currentCalendar().dateByAddingUnit(
+        .Day,
```
wilsonpage: If this is the only thing that differs between Debug and Production, let's just have one function call and pass different value for first argument.

ios/Magnet/AppDelegate.m
```diff
@@ -42,15 +42,12 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
...
-  // when we go to background
-  self.locationReceiver = [[LocationChangeReceiver alloc] init];
-  [self.locationReceiver startSignificantLocationChanges];
+  
+  self.regionManager = [[RegionManager alloc] init];
   
   // Remote logs
   #if DEBUG
-    [Bugfender enableAllWithToken:@"<REPLACE_WITH_APP_KEY>"];
+    [Bugfender enableAllWithToken:@""];
```
wilsonpage, line 50: How does this get the API key?

ios/NotificationsHelper.swift
```diff
@@ -62,10 +62,13 @@ import UserNotifications
   // Throttles the notification process, waiting for 10 seconds until
   // setting up the badge with the number of elements nearby.
   class func notifyUser(url: String, channel: String?) {
-    guard History.getInstance().getRecent(url) == nil else { return }
-    guard toNotify[url] == nil else { return }
+    guard toNotify[url] == nil else {
+      Log.l("Discarding \(url) because is on the list to urls to notify")
```
wilsonpage: `Abort, notification already scheduled for \(url)`

ios/NotificationsHelper.swift
```diff
@@ -62,10 +62,13 @@ import UserNotifications
...
+      return
+    }
     
     if channel != nil && !subscriptions.exists(channel!) {
+      Log.l("Discarding \(url) because we are not subscribed to channel \(channel!)")
```
wilsonpage: `Abort, user not subscribed to channel \(channel)`

ios/NotificationsHelperIOS10.swift
```diff
@@ -38,8 +39,9 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
               subtitle: "by \(channel)",
               body: json[0]["description"].string!,
               url: url)
+        Log.l("Dispatching rich notification for \(json.rawString())")
       } catch {
-        debugPrint("Could not launh notification for \(url) : \(channel)")
+        Log.w("Could not launh notification for \(url) : \(channel)")
```
wilsonpage: `launch`

ios/NotificationsHelperIOS10.swift
```diff
@@ -68,8 +70,15 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
     let category = UNNotificationCategory(identifier: NotificationsHelperIOS10.CATEGORY, actions: [action], intentIdentifiers: [], options: [])
     UNUserNotificationCenter.currentNotificationCenter().setNotificationCategories([category])
     
-    let request = UNNotificationRequest(identifier: url, content: content, trigger: nil)
-    UNUserNotificationCenter.currentNotificationCenter().addNotificationRequest(request, withCompletionHandler: nil)
+    let trigger = UNTimeIntervalNotificationTrigger(timeInterval: 1, repeats: false)
```
wilsonpage, line 74: What's this `trigger` thing? Maybe needs comment?

ios/NotificationsHelperIOS10.swift
```diff
@@ -68,8 +70,15 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
...
+    
+    let request = UNNotificationRequest(identifier: url, content: content, trigger: trigger)
+    UNUserNotificationCenter.currentNotificationCenter().addNotificationRequest(request, withCompletionHandler: {error in
+      guard let error = error else {
+        return
+      }
```
wilsonpage, line 80: Ew, what a waste of space for a `guard` 😝 
Simple `gaurd` one liner lets reader focus on actual function body :)

#theongoingbattle #lol

ios/RegionManager.swift
```diff
@@ -0,0 +1,80 @@
...
+    guard isEnabled else {
+      Log.w("Region monitoring not enabled for this device")
+      return;
+    }
+    
+    scanner = MagnetScanner(callback: self.onItemFound)
+  }
+  
+  private func onItemFound(item: Dictionary<String, AnyObject>) {
+    Log.l("Got item cause region abandoned \(item)")
```
wilsonpage: I find this wording confusing., don't think 'abandoned' is the right word. Perhaps: 'found item on leaving region'.

ios/RegionManager.swift
```diff
@@ -0,0 +1,80 @@
...
+    let url = item["url"] as! String
+    var channel: String? = nil
+    if item["channel_id"] != nil {
+      channel = item["channel_id"] as! String
+    }
+    NotificationsHelper.notifyUser(url, channel: channel)
+  }
+  
+
```
wilsonpage: NIT: Double line break

ios/RegionManager.swift
```diff
@@ -0,0 +1,80 @@
...
+    guard isEnabled && self.currentRegion != nil else {
+      return
+    }
+    Log.l("Stoping listening to region changes")
+    
+    self.locationManager.stopMonitoringForRegion(self.currentRegion!)
+  }
+  
+  private func setupRegion(callback:((Void) -> Void)?) {
+    locationResolver.start { location in
```
wilsonpage, line 74: `.start` suggest the process is ongoing and that there should be a `.stop` isn't this really just a `.get`?

ios/RegionManager.swift
```diff
@@ -0,0 +1,80 @@
...
+      Log.l("Setting up region \(region)")
+      self.locationManager.startMonitoringForRegion(region)
+      if callback != nil {
+        callback!()
+      }
+    }
+  }
+  
+  func locationManager(manager: CLLocationManager, didExitRegion region: CLRegion) {
+    setupRegion(nil)
```
wilsonpage: Is `setupRegion` ever called with a `callback`?

ios/RegionManager.swift
```diff
@@ -0,0 +1,80 @@
...
+    Log.l("Waking up case region abandoned")
+    self.scanner!.start()
```
wilsonpage, line 85: When is `scanner.stop` called if ever?

ios/RegionManager.swift
```diff
@@ -0,0 +1,80 @@
...
+//
+//  Created by Francisco Jordano on 09/11/2016.
+//  Copyright © 2016 Mozilla. All rights reserved.
+//
+
+import Foundation
+import CoreLocation
+import MagnetScannerIOS
+
+@objc(RegionManager) class RegionManager: NSObject, CLLocationManagerDelegate {
```
wilsonpage, line 23: Could you give a high-level explanation of how this class works?

wilsonpage:
r+ after nits

wilsonpage:
Ooopsie sorry

ios/Magnet/AppDelegate.m
arcturus: So far I'm adding it manually when i want to do a build with logging.

Will do a follow up to add the key based on debug type.

ios/RegionManager.swift
arcturus: If we go for get, we will need to have an abort.

Because having the location is not synchronous and we want to be sure we stop getting the location as soon as possible to save battery.

ios/RegionManager.swift
arcturus: No, not necessary, I added at the beginning in case we use this region manager for other things that require callback in the foreground.

But we can remove it and add it back if necessary.

ios/RegionManager.swift
arcturus: Yup, good point.

ios/History.swift
arcturus: Good point

ios/NotificationsHelperIOS10.swift
arcturus: You can trigger notifications with some delays or at a specific date.

Will comment on single line, thought it was pretty readable ... oh gosh, maybe i'm getting used to swift .... is that ok?

ios/NotificationsHelperIOS10.swift
arcturus: it's so clear, it's zen :) also can quickly check nothing else is executed if the guard is not fulfilled :)

ios/RegionManager.swift
arcturus: Sounds good.

ios/RegionManager.swift
arcturus: Never, we don't know if we will find something or when.

Can add a follow up to stop the scanner after 2 minutes, do you agree?

PR #342, opened 2016-11-17
Notifications service extension

arcturus:
Now loading images for notifications.

@wilsonpage this PR includes the commits for the previous PR with new geofencing.

r?

### DIFF
--- a/ios/HeroNotifications/Base.lproj/MainInterface.storyboard
+++ b/ios/HeroNotifications/Base.lproj/MainInterface.storyboard
@@ -1,0 +1,48 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="11542" systemVersion="16B2555" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="M4Y-Lb-cyx">
+    <device id="retina4_7" orientation="portrait">
+        <adaptation id="fullscreen"/>
+    </device>
+    <dependencies>
+        <deployment identifier="iOS"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="11524"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
+    </dependencies>
+    <scenes>
+        <!--Notification View Controller-->
+        <scene sceneID="cwh-vc-ff4">
+            <objects>
+                <viewController id="M4Y-Lb-cyx" userLabel="Notification View Controller" customClass="NotificationViewController" customModule="HeroNotifications" customModuleProvider="target" sceneMemberID="viewController">
+                    <layoutGuides>
+                        <viewControllerLayoutGuide type="top" id="Ft6-oW-KC0"/>
+                        <viewControllerLayoutGuide type="bottom" id="FKl-LY-JtV"/>
+                    </layoutGuides>
+                    <view key="view" contentMode="scaleToFill" simulatedAppContext="notificationCenter" id="S3S-Oj-5AN">
+                        <rect key="frame" x="0.0" y="0.0" width="320" height="240"/>
+                        <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                        <subviews>
+                            <imageView clipsSubviews="YES" userInteractionEnabled="NO" contentMode="scaleToFill" horizontalHuggingPriority="251" verticalHuggingPriority="251" fixedFrame="YES" translatesAutoresizingMaskIntoConstraints="NO" id="zm4-Q4-ugI">
+                                <rect key="frame" x="0.0" y="0.0" width="320" height="300"/>
+                                <autoresizingMask key="autoresizingMask" widthSizable="YES" heightSizable="YES"/>
+                                <color key="backgroundColor" red="0.98862379789352417" green="1" blue="0.99186557531356812" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                            </imageView>
+                        </subviews>
+                        <color key="backgroundColor" red="0.45882353186607361" green="0.74901962280273438" blue="0.66666668653488159" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+                    </view>
+                    <extendedEdge key="edgesForExtendedLayout"/>
+                    <nil key="simulatedStatusBarMetrics"/>
+                    <nil key="simulatedTopBarMetrics"/>
+                    <nil key="simulatedBottomBarMetrics"/>
+                    <freeformSimulatedSizeMetrics key="simulatedDestinationMetrics"/>
+                    <size key="freeformSize" width="320" height="240"/>
+                    <connections>
+                        <outlet property="imageView" destination="zm4-Q4-ugI" id="Bef-6P-2Xe"/>
+                    </connections>
+                </viewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="vXp-U4-Rya" userLabel="First Responder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="33.600000000000001" y="93.553223388305852"/>
+        </scene>
+    </scenes>
+    <color key="tintColor" red="0.99426919221878052" green="1" blue="0.9958687424659729" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+</document>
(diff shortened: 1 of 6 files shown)


arcturus:
You can forget the first 3 commits, and check just the last one.

ios/HeroNotifications/Base.lproj/MainInterface.storyboard
```diff
@@ -0,0 +1,48 @@
...
+                        <outlet property="imageView" destination="zm4-Q4-ugI" id="Bef-6P-2Xe"/>
+                    </connections>
+                </viewController>
+                <placeholder placeholderIdentifier="IBFirstResponder" id="vXp-U4-Rya" userLabel="First Responder" sceneMemberID="firstResponder"/>
+            </objects>
+            <point key="canvasLocation" x="33.600000000000001" y="93.553223388305852"/>
+        </scene>
+    </scenes>
+    <color key="tintColor" red="0.99426919221878052" green="1" blue="0.9958687424659729" alpha="1" colorSpace="custom" customColorSpace="sRGB"/>
+</document>
```
wilsonpage, line 54: Wow, how is this file generated?

arcturus: It's automatically generated by xcode :S

ios/HeroNotifications/NotificationViewController.swift
```diff
@@ -0,0 +1,57 @@
...
+  
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
+  
+    func didReceiveNotification(notification: UNNotification) {
+      NotificationViewController.notification = notification
+      
+      let notifyTimer = NSTimer.init(
+        timeInterval: 1 ,
```
wilsonpage: NIT `1,`

ios/HeroNotifications/NotificationViewController.swift
```diff
@@ -0,0 +1,57 @@
...
+        target: self,
+        selector: #selector(processNotification),
+        userInfo: nil,
+        repeats: false)
+      NSRunLoop.mainRunLoop().addTimer(notifyTimer, forMode: NSDefaultRunLoopMode)
```
wilsonpage: What is the timer for? Can you explain with a comment?

ios/NotificationsHelperIOS10.swift
```diff
@@ -59,11 +60,14 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
     }))
   }
   
-  private func showRichNotification(title: String, subtitle: String, body: String, url: String) {
+  private func showRichNotification(title: String, subtitle: String, body: String, image: String?, url: String) {
     let content = UNMutableNotificationContent()
     content.title = title
     content.subtitle = subtitle
     content.body = body
+    if let image: String = image! {
```
wilsonpage: Why redefine `image` here? Can't you just check it's not `nil`?

ios/HeroNotifications/NotificationViewController.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    @IBOutlet var imageView: UIImageView!
+  
+    private static var notification: UNNotification?
+  
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
+  
+    func didReceiveNotification(notification: UNNotification) {
```
wilsonpage, line 24: Where/how is `NotificaitonViewController` used, is it some ios magic?

wilsonpage:
@arcturus nice work! Some questions inline. How can I test the patch on device/simulator?

ios/NotificationsHelperIOS10.swift
arcturus: There is a huge diference, the first image might not be a string, and canb e undefined.
The second one is for sure a string as we are using the pattern `if let`, so if we check if it's nil, we still need to unwrap (add the !) when we use it. In the current way we are sure what when we use it will be for sure a string.

ios/HeroNotifications/NotificationViewController.swift
arcturus: It's not magic is configured by xcode, into the xmls it says that is the controller for the view that will be displayed, in this case the view for the extended notification. But yes, happens when you initialize the template in xcode.

arcturus:
For testing the patch you need to apply it, well, once I rebased it and when you have a notification pull down to get the expanded view.

arcturus:
Just rebased the patch should be ready to test, here is a video showing it:
https://www.youtube.com/watch?v=W35CylZbhZY

wilsonpage:
>Just rebased the patch should be ready to test, here is a video showing it:
https://www.youtube.com/watch?v=W35CylZbhZY

'Video is private'

wilsonpage:
>For testing the patch you need to apply it, well, once I rebased it and when you have a notification pull down to get the expanded view.

I can use the simulator to mock location and then what? How do I get the notification to be dispatched?

ios/HeroNotifications/Base.lproj/MainInterface.storyboard
```diff
@@ -0,0 +1,48 @@
...
+    <dependencies>
+        <deployment identifier="iOS"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="11524"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
+    </dependencies>
+    <scenes>
+        <!--Notification View Controller-->
+        <scene sceneID="cwh-vc-ff4">
+            <objects>
+                <viewController id="M4Y-Lb-cyx" userLabel="Notification View Controller" customClass="NotificationViewController" customModule="HeroNotifications" customModuleProvider="target" sceneMemberID="viewController">
```
wilsonpage, line 15: Ah I see here the defined controller.

ios/HeroNotifications/NotificationViewController.swift
```diff
@@ -0,0 +1,57 @@
...
+//  Copyright © 2016 Mozilla. All rights reserved.
+//
+
+import UIKit
+import UserNotifications
+import UserNotificationsUI
+
+class NotificationViewController: UIViewController, UNNotificationContentExtension {
+
+    @IBOutlet var imageView: UIImageView!
```
wilsonpage, line 15: What does this annotation do?

ios/NotificationsHelperIOS10.swift
wilsonpage: So is it the same as this?:

```swift
if (image) {
  content.launchImageName = image!
}
```

ios/NotificationsHelperIOS10.swift
```diff
@@ -51,17 +52,20 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
...
     api.post("metadata", data: urls, callback: ApiCallback(success: { json in
-        callback(json)
+      callback(json)
       }, error: { (err) in
         debugPrint("Could not get metadata for \(url): \(err)")
     }))
   }
   
-  private func showRichNotification(title: String, subtitle: String, body: String, url: String) {
+  private func showRichNotification(title: String, subtitle: String, body: String, image: String?, url: String) {
```
wilsonpage, line 61: Does the `?` indicate it can be undefined/nil?

ios/NotificationsHelperIOS10.swift
wilsonpage: It just looks weird that you redefine the variable only to check that it's defined.

ios/NotificationsHelperIOS10.swift
arcturus: Yes

ios/HeroNotifications/NotificationViewController.swift
arcturus: Is a way linking the ui element named 'imageView' defined in the story board, and that's a way of using it as an UIImageView object in code.

arcturus:
> I can use the simulator to mock location and then what? How do I get the notification to be 
> dispatched?

Yes, launch magnet, go to the background with the app, and change the location to the office 📦

ios/NotificationsHelperIOS10.swift
arcturus: I could use a different variable name like _image, but that pattern is quite used in swift to be sure that you get the value that you want from an Optional.

ios/HeroNotifications/NotificationViewController.swift
wilsonpage: ?

wilsonpage:
Looking at the video there is a blank space for a significant period of time until the image loads. Would it be much work to put a stock IOS loading spinner in place until the image has loaded?

ios/HeroNotifications/NotificationViewController.swift
arcturus: uuups, sorry forgot to reply to this comment.

This timer is needed run in the main thread loop. We need to tell when we want to launch our task, in this case `processNotification` in the main thread.

arcturus:
> Looking at the video there is a blank space for a significant period of time until the image loads. Would it be much work to put a stock IOS loading spinner in place until the image has loaded?

I can try to add it, don't know how long it will take. I propose to work today on that, if I have it ready for the end of today the better, if not, we can land this and add a follow up.

Did you try on the phone?

arcturus:
Hei @wilsonpage finally added the spinner, if the network is super fast we wont even see it, other wise I added a light dark background to the notification with a white spinner until we load the image and we replace spinner for image.

wilsonpage:
Managed to test in simulator, works real nice, wicked!

I submitted a PR against your patch to fix some styling issues to do with spinner color and alignment.

wilsonpage:
r+ after that :)

wilsonpage:
NIT: In Android we call these 'ImageNotification's (instead of `HeroNotificaiton`) might be nice to have consistency in naming here. Took me a while to work out what 'hero' meant :)

arcturus:
I'll open a follow up to rename that, although concepts are not the same, as this is a new target, but definitely I think you are right.

PR #343, opened 2016-11-18
Move analytics tracking into an API [android]

samgiles:
This also means we can remove the bridge at some point.  Will have to be a follow up.
The advantage of this is that checking whether we need to send a tracking event at all is entirely contained in the API.

Adds a `NoopCallback` static instance as a convenience for not caring about the result.
Adds to `.gitignore` for vim `swp` files.

For now it uses the wrapper around the GA lib that the bridge has, but we can extract that code into this repository at a later date.

I did test this using adb:

```SH
$ adb shell setprop log.tag.GAv4 DEBUG
$ adb logcat -v time -s GAv4
```

r? @wilsonpage


samgiles:
~~Argh, I just realised formatting and indenting is gonna be all over the place because of my defaults.~~ Actually wasn't that bad.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiMagnet.java
```diff
@@ -17,5 +17,6 @@ public ApiMagnet(Context context) {
         mount("subscriptions", new ApiSubscriptions(context));
         mount("metadata", new ApiMetadata(context));
         mount("preferences", new ApiPreferences(context));
+        mount("tracking", new ApiAnalytics(context));
```
wilsonpage, line 20: The MagnetApi stuff was originally designed to be an interface to the `magnet-service` REST API. I don't think it should become the native interface for everything.

IMO we should expose the native tracking module via it's own dedicated JS API that matches the native API and keep the `api.*` namespace reserved for magnet-service HTTP calls.

android/app/build.gradle
```diff
@@ -121,7 +122,16 @@ android {
...
     buildTypes {
+        def inputFile = file(new File("../../config.json"));
+        def json = new JsonSlurper().parseText(inputFile.text)
+
+        debug {
+            buildConfigField "String", "GA_TRACKER_ID", "\"" + json.analyticsTrackerId.development + "\""
+        }
+
         release {
+            buildConfigField "String", "GA_TRACKER_ID", "\"" + json.analyticsTrackerId.production + "\""
```
wilsonpage, line 133: Nice solution!

android/app/src/main/java/org/mozilla/magnet/notifications/ReceiverNotificationDelete.java
```diff
@@ -15,11 +19,20 @@
...
+
         clearNotification(context, id);
         sendBroadcast(context);
+
+        Api api = new ApiMagnet(context);
+        api.post(
+            "tracking",
+            ApiAnalytics.createEvent("system", "notification-dismiss"),
+            Api.NoopCallback
+        );
```
wilsonpage, line 35: If we forget about MagnetApi, this could be replaced by something like:

``` java
Tracking track = new Tracking(context);
track.event("system", "notification-dismiss");
```

This exact same API could be exposed to JS.

wilsonpage:
Looks like you've arrived at a nice solution, ♠️ce!

Before landing:
- I don't think we should be exposing this stuff through the `api` interface as it's not HTTP and not a magnet-service API. Sorry if this wasn't clear before :( Let's expose a dedicated `track` native API that can replace the `tracking.js` module we have right now.
- Can we have a follow-up issue to remove dependency on `react-native-google-analytics-bridge`?

wilsonpage:
Also what is this for?

> ```
> $ adb shell setprop log.tag.GAv4 DEBUG
> $ adb logcat -v time -s GAv4
> ```

samgiles:
I've ended up making a massive patch that will remove the bridge.  It's in the `no-bridge` branch.

I disagree that this isn't an HTTP API/magnet service thing. Essentially it is - we just happen to use GA as the backend.   In the future, I see us wanting to have our own analytics backend for tracking beacons discovered, beacon health etc.  It seems to make sense to make use of the API mechanism.  It's nice and generic.  It means when we want to add our own tracker for something, we just create a new `CallableTracker` instance, and start calling stuff in the frontend.

I can see this going either way thought tbf.  I mean we could expose each magnet thing as a library for example, either way I'm happy to compromise, this was just easier to implement.

The shell stuff was to turn on the Analytics logging, and then filter only Google Analytics logs, helps you see what's going on.

samgiles:
Replaced by WIP in #349

PR #349, opened 2016-11-21
Native analytics and removes the react GA bridge

samgiles:
This is a follow up/replaces #343 and includes iOS and Android.

Sorry for the mega patch 😭


android/app/src/main/AndroidManifest.xml
```diff
@@ -6,6 +6,7 @@
     <uses-permission android:name="android.permission.BLUETOOTH" />
     <uses-permission android:name="android.permission.BLUETOOTH_ADMIN" />
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION" />
+    <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
```
wilsonpage, line 9: What's this for?

samgiles: It's required by Google Analytics - it doesn't make a difference to what we already have, the bridge manifest used to specify this and get merged with ours.

build-scripts/ios/gen_build_config.js
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env node
```
samgiles, line 1: This file runs as part of the xcode build process and creates "BuildConfig.h".

ios/Analytics.swift
```diff
@@ -10,17 +10,48 @@ import Foundation
...
+    
+    // TODO: Replace this hardcoded value with a value grabbed from config at build time
+    self.tracker = gai.trackerWithTrackingId("UA-87693297-1")
+  }
   
   func trackEvent(category: String, action: String, label: String?, value: Int64?) {
+    let hit = GAIDictionaryBuilder.createEventWithCategory(
+      category, action: action, label: nil, value: nil)
+  
+    if let _label = label {
```
arcturus, line 69: Should you force the unwrap of label?

build-scripts/ios/gen_build_config.js
arcturus: This is awesome!

build-scripts/ios/gen_build_config.js
```diff
@@ -0,0 +1,14 @@
...
+
+const config = require('../../config.json');
```
arcturus, line 3: If we get this from the config file, we move the problem to not to commit the secrets to the config file, should it be some kind of extra file that people may or not in their system, or environment variable as we will need to do it for travis?

ios/Magnet.app/Headers/BuildConfig.h
```diff
@@ -0,0 +1,5 @@
+#if DEBUG
```
arcturus: I guess we should not commit this file and add it to the gitignore right?

ios/Magnet/AppDelegate.m
```diff
@@ -22,9 +22,9 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   NSURL *jsCodeLocation;
 
   // Load from the server
-  // jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
+   jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
   // Load from local bundle
-  jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
+  //jsCodeLocation = [[NSBundle mainBundle] URLForResource:@"main" withExtension:@"jsbundle"];
```
samgiles: Ahh crap, I keep comitting this change - wonder if it's better to use `#if DEBUG` @arcturus ?

build-scripts/ios/gen_build_config.js
samgiles: Environment variable seems like a decent solution, we can use travis secrets then.

android/app/src/main/java/org/mozilla/magnet/magnetapi/ApiPreferences.java
```diff
@@ -11,14 +11,14 @@
 import java.util.Map;
 import java.util.HashMap;
 
-class ApiPreferences extends Api {
+public class ApiPreferences extends Api {
```
wilsonpage: Why expose this instead of using the `ApiMagnet` interface?

android/app/src/main/java/org/mozilla/magnet/tracking/Analytics.java
```diff
@@ -0,0 +1,161 @@
...
+        analytics.setLocalDispatchPeriod(LOCAL_DISPATCH_PERIOD);
+
+        Tracker tracker = analytics.newTracker(aTrackerId);
+        tracker.enableExceptionReporting(true);
+
+        return tracker;
+    }
+
+    private void doTrack(final Callback aTrack) {
+        mPreferences.get("", new Callback() {
```
wilsonpage: Calling `ApiPreferences` directly looks a bit hacky. Could we do `api.get("preferences", ...)` instead?

ios/Analytics.swift
```diff
@@ -0,0 +1,129 @@
...
+  }
+  
+  func doTrackTiming(category: String, value: Double, name: String, label: String?) {
+    let hit:GAIDictionaryBuilder = GAIDictionaryBuilder.createTimingWithCategory(category, interval: NSNumber(double: value), name: name, label: nil)
+    
+    if let _label = label {
+      hit.set(kGAITimingLabel, forKey: _label)
+    }
+    
+    debugPrint("TRACKING TIMING")
```
wilsonpage: Caps required?

android/app/src/main/java/org/mozilla/magnet/tracking/AnalyticsReact.java
```diff
@@ -0,0 +1,72 @@
...
+
+    @ReactMethod
+    public void trackEvent(ReadableMap aData, final Promise promise) {
+        String category = aData.getString("category");
+        String action = aData.getString("action");
+        String label = null;
+        Long longValue = null;
+
+        if (aData.hasKey("label")) {
+            label = aData.getString("label");
```
wilsonpage, line 38: Won't this just be `null` if the key doesn't exist?

wilsonpage: Making `if` redundant?

lib/utils/tracker.js
```diff
@@ -3,21 +3,14 @@
...
-
-const ga = require('react-native-google-analytics-bridge').default;
 const { AsyncStorage } = require('react-native');
 const debug = require('../debug')('Tracker');
-const config = require('../../config');
 
-const analyticsTrackerId = config.analyticsTrackerId;
-const env = global.__DEV__ ? 'development' : 'production';
-const trackerId = analyticsTrackerId[env];
+const tracker = require('./nativetracker');
```
wilsonpage: Consistency: `native-tracker`

ios/AnalyticsReact.swift
```diff
@@ -0,0 +1,49 @@
...
+import Foundation
+import SwiftyJSON
+
+@objc(AnalyticsReact) class AnalyticsReact: NSObject {
+  private let analytics = Analytics()
+  
+  @objc func trackEvent(data: NSDictionary, resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
+    let json = JSON(data);
+    
+
```
wilsonpage: NIT: Remove line

ios/AnalyticsReact.swift
```diff
@@ -0,0 +1,49 @@
...
+@objc(AnalyticsReact) class AnalyticsReact: NSObject {
+  private let analytics = Analytics()
+  
+  @objc func trackEvent(data: NSDictionary, resolve: RCTPromiseResolveBlock, reject: RCTPromiseRejectBlock) {
+    let json = JSON(data);
+    
+    
+    analytics.trackEvent(json, callback: ApiCallback(success: { result in
+      resolve(result.rawValue)
```
wilsonpage: NIT: Indent

ios/Analytics.swift
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func asBoolean(string: String?) -> Bool {
+  if let _string = string {
+    switch(_string) {
+    case "0": return false
+    case "1": return true
+    case "true": return true
+    case "false": return false
+    default: return false
```
wilsonpage: NIT: indentation

android/app/src/main/java/org/mozilla/magnet/tracking/Analytics.java
```diff
@@ -0,0 +1,162 @@
...
+
+        mTracker.send(hit.build());
+    }
+
+    private static Tracker createTracker(Context aContext, String aTrackerId) {
+        GoogleAnalytics analytics = GoogleAnalytics.getInstance(aContext);
+        analytics.setLocalDispatchPeriod(LOCAL_DISPATCH_PERIOD);
+
+        Tracker tracker = analytics.newTracker(aTrackerId);
+        tracker.enableExceptionReporting(true);
```
wilsonpage, line 132: I remember something about using the device/adnvertising ID so that we can identify users across events and track a 'session'. Is that still happening?

wilsonpage:
r+

Hat of to you sir, I know this has been through several iterations. You must know GA inside out now. So long as we haven't lost any tracking metrics that we previously had, let's go a head and ship this mofo! 🚀

wilsonpage:
Looks like a real clean/solid implementation that will work waaaay into the future and give us all the insights we need 😄 #touchwood

arcturus:
woa! fantastic work man!

android/app/src/main/java/org/mozilla/magnet/tracking/AnalyticsReact.java
samgiles: ~~I thought it threw an exception, I'll try it out.~~  It throws an exception if the key doesn't exist so the checks are required.

ios/Analytics.swift
samgiles: Argh, I meant to remove these.

android/app/src/main/java/org/mozilla/magnet/tracking/Analytics.java
samgiles: Yeh, that was a change I made in the Bridge code, I haven't changed any of the defaults from that in here.

config.json
```diff
@@ -0,0 +1,65 @@
...
+
+  "userFlags": {
+    "sortByDistance": {
+      "value": false,
+      "title": "Sort by distance"
+    },
+
+    "showDistance": {
+      "value": false,
+      "title": "Show distance"
```
arcturus: What happened to the config.js file, is still there? Could not see it moved? I think the review ui used to say when a file was renamed.

Anyway, please take a look to:
https://github.com/mozilla-magnet/magnet-client/blob/master/config.js#L17

We landed some support for environment on the userflags keys.

PR #352, opened 2016-11-22
Moving db to an independant file.

### LINKED ISSUE #306: Magnet app for iOS freezes when subscribing/unsubscribing channels
Magnet app for iOS freezes when subscribing/unsubscribing channels. You need to kill the app or stop the simulator and run the app again to make it work again.

Steps:
1. Launch Magnet application in iOS
2. Enter in the Subscriptions menu and subscribe/unsubscribe the channels (It's easy to reproduce, just need to do it 20-25 times)

SAw this error in the traces from the Xcode:

[Issue_subscriptions.txt](https://github.com/mozilla-magnet/magnet-client/files/571922/Issue_subscriptions.txt)

fatal error: 'try!' expression unexpectedly raised an error: database is locked: file /Library/Caches/com.apple.xbs/Sources/swiftlang_swift_2_2/swiftlang_swift_2_2-800.10.12/src/swift/stdlib/public/core/ErrorType.swift, line 54
2016-11-04 16:17:31.875164 Magnet[468:94356] fatal error: 'try!' expression unexpectedly raised an error: database is locked: file /Library/Caches/com.apple.xbs/Sources/swiftlang_swift_2_2/swiftlang_swift_2_2-800.10.12/src/swift/stdlib/public/core/ErrorType.swift, line 54
(lldb)

arcturus:
Also adding locking and transactions to ensure sanity.

Fixes #306

### DIFF
--- a/ios/Subscriptions.swift
+++ b/ios/Subscriptions.swift
@@ -9,6 +9,7 @@ import Foundation
 
 class Subscriptions: NSObject {
   private var db: SubscriptionsDB!
+  private let lockQueue = dispatch_queue_create("com.mozilla.magnet.subscriptions.access", nil)
   
   override init() {
     super.init();
@@ -16,22 +17,42 @@ class Subscriptions: NSObject {
   }
   
   func add(channelName: String) -> Bool {
-    return db.add(channelName)
+    var opResult = false
+    dispatch_sync(lockQueue) {
+      opResult = db.add(channelName)
+    }
+    return opResult
   }
   
   func remove(channelName: String) -> Bool {
-    return db.remove(channelName)
+    var opResult = false
+    dispatch_sync(lockQueue) {
+      opResult = db.remove(channelName)
+    }
+    return opResult
   }
   
   func update(channelName: String, updates: Dictionary<String,AnyObject>) -> Bool {
-    return db.update(channelName, updates: updates)
+    var opResult = false
+    dispatch_sync(lockQueue) {
+      opResult = db.update(channelName, updates: updates)
+    }
+    return opResult
   }
   
   func get() -> [SubscriptionRecord] {
-    return db.get();
+    var subscriptions: [SubscriptionRecord] = []
+    dispatch_sync(lockQueue) {
+      subscriptions = db.get();
+    }
+    return subscriptions
   }
   
   func exists(channelName: String) -> Bool {
-    return db.exists(channelName)
+    var exists = false
+    dispatch_sync(lockQueue) {
+      exists = db.exists(channelName)
+    }
+    return exists
   }
 }
--- a/ios/Subscriptions.swift
+++ b/ios/Subscriptions.swift
@@ -9,7 +9,6 @@ import Foundation
 
 class Subscriptions: NSObject {
   private var db: SubscriptionsDB!
-  private let lockQueue = dispatch_queue_create("com.mozilla.magnet.subscriptions.access", nil)
   
   override init() {
     super.init();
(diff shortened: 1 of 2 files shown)


arcturus:
@sgiles r?

ios/Subscriptions.swift
```diff
@@ -9,29 +9,50 @@ import Foundation
 
 class Subscriptions: NSObject {
   private var db: SubscriptionsDB!
+  private let lockQueue = dispatch_queue_create("com.mozilla.magnet.subscriptions.access", nil)
```
wilsonpage: Can you comment on why the locks are required?

ios/SubscriptionsDB.swift
```diff
@@ -49,48 +55,72 @@ class SubscriptionsDB {
...
-  
-    do { try db.run(insert) }
-    catch { return false }
+    do {
+      try db.transaction(block: {
+        let insert = self.table.insert(
+          self.columnChannelName <- channelName,
+          self.columnNotificationsEnabled <- true)
+        
+        try self.db.run(insert)
```
wilsonpage, line 67: What's the change here? Can you comment to explain :)

wilsonpage:
@arcturus can you explain what was the cause of the issue?

arcturus:
> @arcturus can you explain what was the cause of the issue?

Sorry for not appending the explanation, did an update offline to @samgiles and forgot to add it here.

The bigger problem came because history and channels where using the same sqlite file. In an environment like the office, where we have several beacons we are constantly updateing the history database, cause we find and update the timestamp constantly. So when you use the channel you can have situations where two different threads class trying to write to same db, but different tables.

Now each component writes on its own db, also added transactions and locking, to be sure we don't have any problem when writing on the DBs.

ios/Subscriptions.swift
arcturus: Just to make sure we make the operations atomic.

ios/SubscriptionsDB.swift
arcturus: I'm wrapping all the operations into transactions, it's a standard way of working with dbs. If for any reason the code between the transaction fails, the DB will rollback any change made to the db and we wont have an inconsistent state.

ios/SubscriptionsDB.swift
```diff
@@ -37,7 +36,14 @@ class SubscriptionsDB {
       true).first!
     
     // get handle on database
-    db = try! Connection("\(path)/history.sqlite3")
+    db = try! Connection("\(path)/subscriptions.sqlite3")
+    db.busyTimeout = 5
+    db.busyHandler({ tries in
+      if tries >= 3 {
```
samgiles: NIT: Magic numbers 3 and 5!

ios/Subscriptions.swift
samgiles: In this file is it necessary to lock _and_ have transactions in the other file?  It looks odd in here that we lock around operations that are supposedly atomic?

If the extra lock was necessary, can we move them into `SubscriptionsDB` (and then use only one lock in there for initialisation and operations)?

ios/SubscriptionsDB.swift
```diff
@@ -37,7 +36,14 @@ class SubscriptionsDB {
       true).first!
     
     // get handle on database
-    db = try! Connection("\(path)/history.sqlite3")
+    db = try! Connection("\(path)/subscriptions.sqlite3")
```
samgiles: NIT: Can we move this to a constant too?

samgiles (commented):
Looks good to me! Just a couple of comments -  🚢!

ios/SubscriptionsDB.swift
arcturus: lol, ok will move them to constants you are right :P

arcturus:
> In this file is it necessary to lock and have transactions in the other file? It looks odd in here that we lock around operations that are supposedly atomic?
> 
> If the extra lock was necessary, can we move them into SubscriptionsDB (and then use only one lock in there for initialisation and operations)?
> 

Fair enough, I'll give it a try without the looking now that we have it in a separate file. Sure, makes sense just with transactions we should not have this error.

arcturus:
Just tried out without the locking, just transactions and works smoothly, so, once travis gives me a green will merge.

samgiles:
Nice! 🏆

PR #355, opened 2016-11-23
Deep linking to item view when notification is clicked

arcturus:
@wilsonpage r?

Before we were just handling the click on the custom actions, but not the normal notification.

This will end fixing #189

### DIFF
--- a/ios/NotificationsHelperIOS10.swift
+++ b/ios/NotificationsHelperIOS10.swift
@@ -85,6 +85,11 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
       debugPrint("Launching web page \(response.notification.request.identifier)")
       let url = NSURL(string: response.notification.request.identifier)
       UIApplication.sharedApplication().openURL(url!, options: [:], completionHandler: nil)
+    } else {
+      // We were clicked but not the visit (go to browser) action
+      // so we navigate to the deep link inside the app
+      let url = NSURL(string: "mozilla-magnet:item?url=\(response.notification.request.identifier)")
+      UIApplication.sharedApplication().openURL(url!, options: [:], completionHandler: nil)
     }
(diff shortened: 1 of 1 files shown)


ios/NotificationsHelperIOS10.swift
```diff
@@ -85,6 +85,11 @@ class NotificationsHelperIOS10: NSObject, UNUserNotificationCenterDelegate {
       debugPrint("Launching web page \(response.notification.request.identifier)")
       let url = NSURL(string: response.notification.request.identifier)
       UIApplication.sharedApplication().openURL(url!, options: [:], completionHandler: nil)
+    } else {
+      // We were clicked but not the visit (go to browser) action
+      // so we navigate to the deep link inside the app
+      let url = NSURL(string: "mozilla-magnet:item?url=\(response.notification.request.identifier)")
```
wilsonpage: Shouldn't this be: `mozilla-magnet://item?url=...`

wilsonpage:
- The custom protocol looks wrong, it's missing the `//`
- Shouldn't we also be deep linking for < iOS-10?

wilsonpage:
r+ after those comments addressed

ios/NotificationsHelperIOS10.swift
arcturus: Yeah you are right, it's working, but definitely better to add the //

arcturus:
> Shouldn't we also be deep linking for < iOS-10?

Yes, thanks for the remind, it's pretty easy to do in the AppDelegate.
Good catch!

arcturus:
Oh I just realised, for ios9 we say 'Content nearby' we don't show specific notifications for poi, so with that behaviour makes sense to send the user to the list. 

As well we don't know what the user wants to see, as that notification could mean that you found several things.

PR #359, opened 2016-11-24
For some reason 'setValue' on a `GAIDictionaryBuilder` doesn't work correctly in iOS

samgiles:
This is a somewhat hacky workaround to get it working.

r? @arcturus


arcturus:
Tested on device, working pretty well, thanks @samgiles !

PR #364, opened 2016-11-24
Better handling of the errors that might happend in the api

### LINKED ISSUE #361: Crash opening the app when wifi&data are off  and there are physical beacons around
STR:

0. Magnet app is not opened and we have, at least, a physical beacon around
1. Deactivate Data or switch the Airplane mode on
2. Open Magnet app
Expected Result: Network error
Current Result: Red screen shown
img_0035

arcturus:
Fixes #361 

r=me
r= @oteo


arcturus:
Merging after @oteo checked

PR #376, opened 2016-11-24
Remove swipe logic

wilsonpage:


wilsonpage:
@arcturus r? Let's land this after the release, just in-case :)

Issue #379, opened 2016-11-25
Telemetry - Telemetry setting needs to be turned off and on to start collecting data


isabelrios:
Something is not going on well with telemetry. It seems that now it is necessary to disable and enable telemetry setting and close and open the app to have the data recorded in GA. 
I do not remember having to do that in order to get the data recorded in the past.

Tested with:
Dev builds, nightly and the released build.

oteo:
@samgiles or @wilsonpage can you have a look at it, please?

PR #382, opened 2016-11-25
Ensure default preferences are always exposed

samgiles:
iOS patch for #379 

r? @wilsonpage

### DIFF
--- a/ios/Magnet/AppDelegate.m
+++ b/ios/Magnet/AppDelegate.m
@@ -23,9 +23,9 @@
   NSURL *jsCodeLocation;
 
   // Load from the server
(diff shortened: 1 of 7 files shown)


ios/Magnet/AppDelegate.m
```diff
@@ -23,7 +23,7 @@ - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(
   NSURL *jsCodeLocation;
 
   // Load from the server
-  // jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
+  //jsCodeLocation = [[RCTBundleURLProvider sharedSettings] jsBundleURLForBundleRoot:@"index.ios" fallbackResource:nil];
```
arcturus, line 26: Not needed, but we wont block on this one :)

PR #383, opened 2016-11-25
Hide sort by distance (only visible in debug mode)

### LINKED ISSUE #381: Make settings: Sort by distance only available for debug builds
As there are issues with "Sort by distance" functionality (issue #367) let's do this setting only available for debug builds as we already did in issue #337 for Show distance and Remove old items.

samgiles:
Closes #381

r? @wilsonpage

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,7 +8,8 @@
   "userFlags": {
     "sortByDistance": {
       "value": false,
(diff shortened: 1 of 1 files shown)


arcturus:
r=me

